Issue #643, opened 2014-05-13
Modify TinyXML to use FileStream


lukaspj:
We should modify TinyXML to use FileStream as they have done in T2D, so that it is compatible with T3D's virtual file system (e.g. reading files inside zip files)
As pointed out by TRON in IRC:

> it won't work with the virtual file system, therefore things like storing files that library tries to open in zip archives would fail.

Breaks the "lib"-properties of TinyXML, but it should be done imo.

Issue is that collada_dom relies on TinyXML so integrating it directly into T3D makes collada_dom rely on the T3D DLL which is bad practice, but I guess collada_dom should use the virtual file system as well?

lukaspj:
Should be changed to "modify use of TinyXML to avoid using a modified version" as these changes were merged in later on

PR #651, opened 2014-05-14
Make TinyXML and SimXMLDocument use the virtual file system.

lukaspj:
As described in issue #643, this pull requests aims to modify the TinyXML library in a way that doesn't affect current behaviour, but makes it possible to use FileStreams for working with the xml files, which is necessary to work with T3D's virtual file system (e.g. works inside Zip folders and stuff like that)

This Pull Request has some issues (such as placement of files and not being able to automatically merge) which I'll fix if we deem it is useful to T3D.

It's split into 4 commits: 
* dd4ae89 fixes some inconsistencies between T2D and T3D in the Stream objects.
* 50c65b5 Makes the TinyXML library more inheritance friendly.
* f40d50a Adds FileStream versions of the TinyXML classes.
* ba9bccf Alters the the SimXMLDocument classes to make use of the virtual file system through the FileStream versions of the TinyXML classes.

They are dependent on each other from bottom to top.

So is this a change we want in T3D? Discuss.

### DIFF
--- a/Engine/lib/tinyxml/tinyxml.h
+++ b/Engine/lib/tinyxml/tinyxml.h
@@ -283,10 +283,8 @@ public:
 		TIXML_ERROR_STRING_COUNT
 	};
 
-protected:
-
 	static const char* SkipWhiteSpace( const char*, TiXmlEncoding encoding );
-
+protected:
 	inline static bool IsWhiteSpace( char c )		
 	{ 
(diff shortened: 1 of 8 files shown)


Lopuska:
I close this as we think it's better to don't customize external libraries. However the feature is really useful so we will consider again this PR when we will have a version which allow file system customization without interfering with internal code library.
Also this will be also the right moment to take a look on TinyXML2.
Thank you! (:

PR #676, opened 2014-05-30
Fix dedicated Linux build.

LuisAntonRebollo:
Work OK on Ubuntu and Windows 7.


Duion:
Is this tested and works on a server box, not a desktop? I have openSuse and I can't compile the dedicated server and the regular one run as dedicated crashes with "Format GFXFormatB8G8R8A8 not supported with specified profile.
% Segfault" when trying to launch it.

PR #708, opened 2014-06-24
Alpha Skyboxes that can use lighting

xoltan:
Both types of cloud layers have a setting to appear in front of or behind the skybox.  Normal skyboxes need no changes.  Alpha skybox material needs transparency checked in the material editor.


Lopuska:
I'll try to get this working and bug free.
Thanks for contribution!

PR #710, opened 2014-06-30
3 textures blending fix

lukaspj:
While working with the terrain blending, I found some artifacts when having more than 2 textures blended together.
Steve_Yorkshire in IRC confirmed that he had also seen the issue, so I dug into it and found that the textures aren't blended properly. This PR fixes specifically the blending between the 2nd and the third texture.

Image of the issue here:
Blend artifacts

With this PR it looks like this:
Blend fixed

### DIFF
--- a/Engine/source/terrain/terrRender.cpp
+++ b/Engine/source/terrain/terrRender.cpp
@@ -170,9 +170,10 @@ bool TerrainBlock::_initBaseShader()
    desc.zDefined = true;
    desc.zWriteEnable = false;
    desc.zEnable = false;
-   desc.setBlend( true, GFXBlendSrcAlpha, GFXBlendInvSrcAlpha );
+   desc.setBlend( true, GFXBlendSrcAlpha, GFXBlendOne );
    desc.cullDefined = true;
    desc.cullMode = GFXCullNone;
+   desc.colorWriteAlpha = false;
    mBaseShaderSB = GFX->createStateBlock( desc );
 
    return true;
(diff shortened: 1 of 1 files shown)


lukaspj:
@RichardsGameStudio It's been a while, but in case you are still available, could you have a look at #920? I just added a new commit with some new changes in it. I've not been able to reproduce the seams after those changes.

RichardsGameStudio:
@lukaspj Sorry for the late reaction, but it got into the pile of things I had to do and it was growing faster than I realized. I will try to find some time this weekend to test it.

eightyeight:
That'd be great :).

RichardsGameStudio:
@lukaspj I've tried the suggested solution, deleted all the procedural shaders, deleted the basetex file, but still the same fuzzy borders. I am using terrainmaterials with all the maps set (diffuse, macro, detail and normal).

lukaspj:
@RichardsGameStudio I'd love to get a minimal working example, like if you could reproduce the effect in the empty terrain level and send the whole project to me!

RichardsGameStudio:
@lukaspj I'll see what I can do. The problem is that there are parts of the engine that I have modified which I can not make public, but as said I'll see what I can do for you.

lukaspj:
@RichardsGameStudio I don't need the actual source code.. Just like, a mission with a flat terrain and textures which can be used to reproduce the issue. I'd use the stock repo to test the mission anyway, as I only need to test the terrain rendering :)

PR #714, opened 2014-07-02
Editor integration for new blending

lukaspj:
Added GUI controls for enabling / disabling the new terrain blending and modifying the blendDepth as described in #712 .

[Watch the new controls in action here](https://www.youtube.com/watch?v=jTawTt4Lkv0)


Lopuska:
@lukaspj Hello! (:
Thanks again for this nice work.
I'm working on this, especially to profile it and evaluate the render cost.
I see a for loop in the shader that could be unrolled since the iteration number is known at compile time. Other improvement could be done.
I've also already ported it to OpenGL.
I'll give updates progressively and make probably a new PR, for the final one.
Thank you!

PR #719, opened 2014-07-09
Fix to allow parallax mapping with dxtnm textures via the red channel.

rextimmy:
Currently the only option in T3D to use compressed normal maps with parallax is DXT3. This allows allows the use of height in the red channel of a dxt5nm.

A test texture set is here http://narivtech.com/downloads/textures_normal.zip


eightyeight:
Does this need an OpenGL equivalent? Going to have to push it back, for now, sorry :(.

Azaezel:
yes, needs an opengl equivalent. forgot that when I closed https://github.com/GarageGames/Torque3D/pull/849

PR #746, opened 2014-08-02
New SSAO method

lukaspj:
As described [here](http://www.garagegames.com/community/blogs/view/21060), this change will give both an improvement in the visual results of SSAO as well as a significant FPS improvement.

**However,** it should be noted that this PR doesn't remove the existing SSAO method, and the only way to change between them is to rename the original "SSAO.cs" to something Torque doesn't find like "SSAO.ncs".

This PR is mostly here for discussion and I will change it to a PR that replaces the current SSAO implementation if there is an interest. Personally I've had a lot more success with the new SSAO method.


lukaspj:
It should now completely replace the old one.

eightyeight:
The CI errors are all:

    ../../../../Engine/source/sfx/openal/LoadOAL.h:34:12: fatal error: 'al/al.h' file not found
    #  include <al/al.h>
               ^

Which is perplexing.

lukaspj:
Made a small change to `Templates/Full/game/core/scripts/client/postFx/default.postfxpreset.cs` but I doubt that triggered the bug.. I've no clue...

VinnyVicious:
Has this been abandoned? @lukaspj, @eightyeight

PR #750, opened 2014-08-06
Particle system refactor

lukaspj:
As described [here](http://www.garagegames.com/community/blogs/view/22801).

I believe I've managed to port all the existing features to the new Particle System except for the "multiple particledatablocks per system" feature, which I'm still not sure of the purpose of.
It could possibly be created by having a "compositeParticleSystem" that creates 2 ParticleSystems and emits from both of them.

This has only been lightly tested and heavy testing before merging is recommended.

Things that especially need testing:
- Animated billboard textures.
- Integration into existing classes.

If this PR goes through, then I will be releasing some of my IPS documentation for the ParticleSystem.


VinnyVicious:
Has this been abandoned?

John3:
@VinnyVicious AFAIK this and mayor changes will be include in the next version v4.0 due backward compatibility.

PR #815, opened 2014-10-05
Add additional padding on unescape scenarios.  Also fix appendformat whi...

figment:
expandEscapes was causing corruption on strings as worst case scenario was not worse enough.  

String::StrFormat::formatAppend was not using vsnprintf correctly and would corrupt when crossing allocation boundaries when appending to an existing string.

### DIFF
--- a/Engine/source/core/util/str.cpp
+++ b/Engine/source/core/util/str.cpp
@@ -1334,7 +1334,7 @@ String String::trim() const
 
 String String::expandEscapes() const
 {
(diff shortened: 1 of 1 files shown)


Lopuska:
Hello, thanks for contribution.
I'm unable to reproduce this issue, also magic numbers are not recommended as they aren't enough descriptive and doesn't give you the guarantee that this additional padding of 32 will be enough.
It's better to fix this in a different proper way if will get this kind of crash again.

PR #863, opened 2014-10-24
glow buffer debug visualiser

Azaezel:
pure script and shader, there will be several of these using identical methodology for deferred. displays a given buffer via the
 texture[0] = "#glowbuffer"; entry,
passed along to a given shader via the
samplerNames[0] = "glowBuffer";entry
which corresponds to an expected variable name and register.

as much documentation for later as tool


eightyeight:
Thanks for the new tool :). I should make a note to port this to the Empty template as well, though I still aim to get rid of the duplicated templates anyway so I'm not incredibly fussed.

PR #896, opened 2014-11-04
Dev   forest wind emitter improvement

Winterleaf:
Fixes the sphere of red from disappearing when the camera is inside the range of the wind.


eightyeight:
Revising this to remove the two `Point3F` utility methods.

PR #897, opened 2014-11-04
Minor Sanity Check for power of two on textures.

Winterleaf:
Just a power of two check


eightyeight:
I'm inclined to leave this one out. @Areloch?

Areloch:
Yeah, I think so.

Just writing to the console doesn't make it immediately apparent, and also would spam the console of the end-game, even if the artists decided that non-pow2 textures were legitimately what they wanted(for whatever reason).

Ideally, this is the sort of thing we'd implement at the tool level, so if you use a non-pow2 texture in a material for instance, it'd have a pop up, or warning flag, etc.

eightyeight:
Sorry @Winterleaf :(.

PR #901, opened 2014-11-04
FFP 64bit

Winterleaf:
...px

### DIFF
--- a/Engine/source/platformWin32/winMath_ASM.cpp
+++ b/Engine/source/platformWin32/winMath_ASM.cpp
@@ -91,25 +91,27 @@ void Platform::setMathControlStateKnown()
       fldcw cw
    }
 }
-
 #else
-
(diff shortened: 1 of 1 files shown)


LuisAntonRebollo:
Renamed and Moved to 3.8 for disscuss.

TheDushan:
Direct3D will set the floating-point unit to single-precision, round-to-nearest as part of initialization for the calling thread, unless the D3DCREATE_FPU_PRESERVE flag is used, in which case, the floating-point control word is untouched
Since the control word is a per-thread setting, ensuring that all of your application threads are set to single-precision mode may optimize performance.
x86-64 native code never uses the FPU, and always uses SSE instructions, so the floating-point control word issues are all specific to x86 code,  _control_fp is not valid for x64-native coding, which uses SSE exclusively instead.

Azaezel:
this one seems to check out

PR #904, opened 2014-11-05
Converted the file system console methods to be engine-based

Areloch:
Converted the file system console methods to be engine-based that the console functions call to. This doesn't impact usage or functionality, but enables the functions to be used in the engine as well.


eightyeight:
@Areloch I think we should leave this out of 3.8. Needs fixing but it's not critical.

Areloch:
Yeah, we can bump it to 3.9.

Areloch:
This needs to be re-implemented.

PR #909, opened 2014-11-06
Client side Far Clipping adjuster, Client side scene culling and fixing the hitch when players zoom out after loading in.

Winterleaf:
I couldn't create separate pulls due to the way these features are inter-dependant.

First, I've added Far Clipping Client culling to the scene.  In my benchmarks it increases framerate by 5-10 fps in heavily populated zones.
I've added a new global TorqueScript Variable $pref::Camera::FarDistance to allow easy setting and getting of the clipping distance.
I also added it to the Options Dialog for in game use.
The max value the Far Clipping Distance can be is the max Viewable distance defined in the mission.

With this patch, I've also included a fix for the hitching for when players zoom out for the first time or when they enter a heavily populated area.
The idea is to load all LOD's of all Shapebase and Static objects into the video card before the first render.  This removes the hitch players
see when they scroll out after loading into the game.

### DIFF
--- a/Engine/source/scene/sceneRenderState.cpp
+++ b/Engine/source/scene/sceneRenderState.cpp
@@ -30,7 +30,7 @@
 //<!-- Scene Culling --!>
 #include "console/engineAPI.h"
 #include "T3D/gameBase/gameConnection.h"
-#include "terrain/terrdata.h"
+#include "terrain/terrData.h"
 
 namespace CameraAndFOV
 	{


Winterleaf:
What is causing the merge conflicts?

eightyeight:
Don't know - merge `development` into this branch and you'll see.

rextimmy:
Hey testing this out i still see heaps of "hiccups" during game-play and the culling is very aggressive!

jojojubi:
Until recently I had this installed in my game. When you switch from first person to third person camera, the near plane is not adjusted to the new camera position, resulting in close objects getting culled out while visible.

Lopuska:
@Winterleaf, @lukaspj
Thanks so much for this!
I'm working on review and clean a lot of PRs which are pending since long time.
This features looks interesting, however I have issues when using it in Third Person mode.
e.g.
Try to use CheetahCar and place some small shape object behind. You'll see these objects disappear before they get out from frustum. Something wrong with plane collision test maybe?
I also have a question: What this frustum cull differs from the the current T3D culler?
T3D shouldn't draw any object that have its bounding box out from camera frustum. (I tryed using frustum lock to debug), but I could be wrong.
Thank you for contribution.

Winterleaf:
I never figured out the issue w/ 3rd person and we have since moved on to UE4.  But the purpose of this patch was to allow players to control there far and near fustrums for performance reasons.  Say the screen has a lot of particle effects, stuff moving etc.  They could shorten the distance the client drew everything to make it play better.

The other purpose was to fix the hitch you get when you first log into a game.  I'm not sure if it still happens in T3D since it's been many years since I messed with it.  But back then, if you had a lot of models in view and the player used the mouse wheel to zoom out, the client would hitch for a couple seconds.  By pre-rendering all the models w/ there LOD in the client before the game built the first screen we were able to remove that initial hitch.

As for the culling, once again, I'm not sure on the current state of things, but back then the culling wasn't very aggressive and it would render things off screen or hidden behind other objects.  Our patch tended to pick up 5-10 fps viewing the same scene with it enabled.

Hope this info helps you get some inside into why it was written.

Engine/source/T3D/tsStatic.cpp
```diff
@@ -494,6 +498,21 @@ void TSStatic::prepRenderImage( SceneRenderState* state )
 {
    if( !mShapeInstance )
       return;
+//<!-- Scene Culling --!>
+   GameConnection* connection = GameConnection::getConnectionToServer();
```
Winterleaf: this is where it did the pre-render of the models in the scene to get rid of the hitch

Engine/source/T3D/tsStatic.cpp
```diff
@@ -494,6 +498,21 @@ void TSStatic::prepRenderImage( SceneRenderState* state )
...
+	if (connection && !connection->didFirstRender)
+   {
+      TSRenderState rdata;
+      rdata.setSceneState( state );
+      rdata.setFadeOverride( 1.0f );
+      rdata.setOriginSort( mUseOriginSort );
+      for (S32 i = mShapeInstance->getSmallestVisibleDL(); i >= 0; i-- )
+      {
```
Winterleaf: We drew each lod to the buffer to preload it before the first render occurred.

Templates/Full/game/art/gui/optionsDlg.gui
```diff
@@ -1312,6 +1312,131 @@
                canSaveDynamicFields = "0";
             };
          };
+         new GuiTextCtrl() {
```
Winterleaf: Added the sliders for adjusting the near and far culling to the settings ui

Lopuska:
Thanks Vince for your detailed explanation.
I'll take a look deeply about the problem I mentioned you.
If I have questions I'll ask you here.

Thank you!!!

Lopuska:
o.o I just noted that @jojojubi got my same bug.

Issue #679, opened 2014-05-31
PostEffect Doesn't clear out constants on shader reload


jamesu:
In the PostEffect class, if you cause the shader to be updated (e.g. by changing a macro via `setShaderMacro()` or `reload()`), the shader will be cleared in [checkRequirements](https://github.com/GarageGames/Torque3D/blob/development/Engine/source/postFx/postEffect.cpp#L1389), but the shader constant buffer (mShaderConsts) and the constants list (mEffectConsts) will not be updated to point to the new shader.

This causes problems in the OpenGL renderer as the constant buffer state becomes out of sync with the shader state. e.g. the render target size may be 0,0 since it still thinks the size is 1024,1024. The Direct3D renderer doesn't appear to be affected by this, possibly because it updates using ranges rather than on a per-constant basis.

eightyeight:
Any update on this?

LuisAntonRebollo:
Moved to 3.8 for get time to confirm/test.

PR #944, opened 2014-11-14
Postfx shutoffs

Azaezel:
Adds:
$pref::Video::disableSSAO = false;
$pref::Video::disableHDR = false;
$pref::Video::disableLightRays = false;
$pref::Video::disableDOF = false;
$pref::Video::disableVignette = false;
a [Highest] entry in the shader quality dropdown.

as new prefs, which can override per map postfx settings
[Highest] = everything on
[High] = shuts off SSAO
[Normal] = additionally disables HDR
[Low] = for new flags is the same as normal, keeps the older shutoffs.
[Lowest] = takes no prisoners.

### DIFF
--- a/Templates/Full/game/core/scripts/client/defaults.cs
+++ b/Templates/Full/game/core/scripts/client/defaults.cs
@@ -333,7 +333,7 @@ new SimGroup( TextureQualityGroup )
       key["$pref::Video::textureReductionLevel"] = 0;
       key["$pref::Reflect::refractTexScale"] = 1.25;
       key["$pref::Terrain::detailScale"] = 1.5;      
-   };   
+   };
 };
 
 function TextureQualityGroup::onApply( %this, %level )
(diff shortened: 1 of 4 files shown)


Templates/Full/game/core/scripts/client/defaults.cs
```diff
@@ -436,11 +457,34 @@ class = "GraphicsQualityLevel";
...
+      key["$pref::Video::disableVignette"] = false;
+      key["$pref::Video::disableDOF"] = false;
+      key["$pref::Video::disableLightRays"] = false;
+   };
 };
 
-
+function ShaderQualityGroup::onApply( %this, %level )
+{
+   PostFXManager.settingsSetEnabled(true);
```
Azaezel: This bit came up for discussion RE: https://github.com/GarageGames/Torque3D/issues/679
intended to work in conjunction with https://github.com/GarageGames/Torque3D/blob/39f0e269d6c4f653f6424c70bc580e64545a9d7d/Templates/Empty/game/core/scripts/client/postFx/postFxManager.gui.settings.cs#L39 among other things to resize the backbuffer *before* you load into game and have to do it anyway per-level.

Azaezel:
Closing this as incompatible and likely redundant due to the template refactor work.

PR #951, opened 2014-11-20
Console part 2 callbacks and auit

Winterleaf:
Ok, this is kinda a big PR.  Because the AUIT and the console callback refactor are so intertwined I wasn't able to separate them into different PR's.

I emailed Dan the user manual for the AUIT.  Winterleaf Entertainment L.L.C. is releasing all the documentation and source code to the MIT project.

Let me know if you find any issues

### DIFF
--- a/Engine/source/util/ReadXML.cpp
+++ b/Engine/source/util/ReadXML.cpp
@@ -10,13 +10,13 @@
 IMPLEMENT_CONOBJECT(ReadXML);
 
 ConsoleDocClass( ReadXML,
-				"@brief Class used for writing out preferences and settings for editors\n\n"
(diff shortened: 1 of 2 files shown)


TheDushan:
There is bug inside tools/guiEditor/scripts/guiEditorCanvas.ed.cs 
line (223): Unknown command delete.

Steps how to reproduce problem.
On main menu, open GUI Editor (F10) choose "ConsoleDlg" and try to exit from GUI Editor,

eightyeight:
Sorry for the lack of progress on this. This may get postponed to 3.7. There are definitely some features we want (multi window), and some that I'm pretty sure we could do without, and I may look into splitting the console changes off a little bit.

PR #959, opened 2014-11-25
Asset downloader

Winterleaf:
This PR provides a mechanism for synchronizing files between the client and the server.
It does allow users to upload content to the server if the file transfer connection is left connected.
The code is implemented using the TCPObject as the base class.  I have taken the liberty to make a couple of TCPObject's functions virtual so I could overload them for binary transfer.
I also modified fileSystemFunctions.cpp and took the body of code out each of the defineconsolefunctions and put them in a function and wrapped all the functions inside a namespace.  I also added a header file which can be included so that programmers are able to access those functions from anywhere in the code base.
Since the FileXfer occurs over TCP,  I am able to use the same port which is used for the game.  This is because T3D runs its network over UDP and this PR is implemented over tcp.

You are able to specify the files you want to include for download by editing the core/Scripts/server/fileXferServer.cs file.

Two ways of adding files to be available for download:

LoadPath (Pattern, Recursive, multiMatch, Verbose) and works just like the FindFiles file searching.
LoadFile (filename) - It will check if the file exists, if it exists it will add it to the download.

Let me know if you find any errors.

Vince Gee
Winterleafentertainment.com

### DIFF
--- a/Engine/source/app/net/netFileClient.cpp
+++ b/Engine/source/app/net/netFileClient.cpp
@@ -169,7 +169,7 @@ bool netFileClient::processLine(UTF8 *line)
 
 bool netFileClient::prepareClientWrite(const char* filename,U32 size)
 {
-   if ( fileSystemFunctions::isWriteableFileName( filename ) )
+   if ( Torque::FileSystem::isWriteableFileName( filename ) )
    {
       xferFile = new FileObject();
       xferFile->openForWrite(filename, false);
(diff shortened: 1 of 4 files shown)


Winterleaf:
Is there something that needs to be fixed on this pull?

eightyeight:
We're locking down 3.7, so I'm moving all the non-bugfix PRs back. Sorry, we've just got a lot of work to do for 3.7. Trying to minimise it :P.

PR #960, opened 2014-11-27
An alternative to PathShape

Winterleaf:
This is from http://www.garagegames.com/community/forums/viewthread/137542

I'm calling this resource WalkableShapes (I'm open to suggestions for better class names). It provides three classes that can be used in games. TSPathShape is a simple path following shape. TSAttachable builds on TSPathShape to add the ability for players to walk on the shape while it's moving. The script interface for this class is almost identical to the traditional PathShape. The setup is slightly different. Since the DataBlock has been eliminated, the shape file is set in the editor. The WalkableShape class is a simple shape object that players can walk on while it's moving. It provides no movement of it's own, but can be mounted to...Anything! So if you want to walk on the back of a flatbed truck, model the bed as a separate shape and mount it to the truck in game as a WalkableShape. It can be mounted to any shape that provides movement. There is a demo level included that has a platform-jumping mini-game and a Cheetah with a platform mounted to the roof where the turret normally mounts. You'll need two connections to try that though.


Lopuska:
What about that? Does someone is interested on this? tsDynamic is a new class we need to merge with the new tsStatic additions since last year. Also, does bullet physics compatible with that? I remember long time ago I had troubles on that because of gravity which still push down also bullet ghostObjects. Please someone can verify that?

Winterleaf:
Hello again!  Ok, this resource allowed you to have moving platforms, I needed it to implement a floating disc game where players would jump disc to disc and I wanted the disc's to move and such.  I did the initail PR but I didn't write it, some other guy did.

Azaezel:
https://github.com/OTHGMars/Torque3D/tree/walkable_shapes the origional

Azaezel:
https://github.com/Azaezel/Torque3D/tree/walkableShapes_Mege contains a merge-updated version, as well as some initial factfinding for working around the reason this was apparently rejected. Turns out with the physics plugins, players would attach, but continue falling. https://github.com/Azaezel/Torque3D/commit/efa2f9abe3b8efcb4ba686b18b1909e9bdafe741#diff-b8335f02a980665dd7dc953db7a637caR2804 Seems to do the job for bullet at least. No guarantees on physX.

*Note: On talking with @Areloch , this may remain sitting in the queue as a reference until such time as the general methodology is rolled in as a component.

PR #973, opened 2014-11-29
Implementing crunch (DXT compression library) - replaces 'squish'.

just-bank:
At the moment this implementation uses fastest compression method, but so far it looks like its better than fastest squish (at least for terrain basemaps).


Lopuska:
Well, we have a problem with that.
It's designed to work only on windows.
It's not a good idea for us to specialize libraries per platform.
Squish works really well, it's designed to be as fast as making sense to do this job at loading time. (we can try to activate SIMD code compile path if we want to gain some speed on it), I don't think it's activated by default.
As alternative, I'd rather recommend a GPU texture compressor to gain seriously better performance. Compute shader stage is really useful for these kind of job.
If someone complains for slowness about squish, just open an issue ticket and we'll try to get a solution.
Does someone has other opinions about that?

Areloch:
Yeah, I think looking forward to a GPU-driven solution is probably a good idea.

I'd say we could move this discussion to an issue so we can try and pick through a good, fast, cross-plat solution. We do appreciate the suggestion though, bank :)

PR #975, opened 2014-11-29
Added support to read PSD files.

just-bank:


Lopuska:
Nice addition, I'm personally for merge this.
After all, Unreal Engine can load *.psd resources if I remember correctly.
The goal of this shouldn't be use this feature as final textures format representation on the game for obvious performance and convenience reasons due to size and usage 	
comfort, but it's useful instead for prototyping the game and during development process.
Artists would love this. Tech addition is never bad.
Someone has any other opinions on that? (:
Thank you!

PR #978, opened 2014-11-29
Support for Physx 3.3.2 with VS 2010

rextimmy:
This one is a tricky one, starting with Physx 3.3.2 NVidia are now shipping separate dll/lib for different versions of Visual Studio. This PR adds support for VS 2010 (which technically is the max version project gen supports) but if you are using a later version of VS it won't compile (you need the matching versions). You would need to open up physx3.inc file and change "vc10" to whatever version you are using (currently NVidia only support vc9/10/11)

I'm really not sure what to do about this?

### DIFF
--- a/Tools/projectGenerator/modules/physX3.inc
+++ b/Tools/projectGenerator/modules/physX3.inc
@@ -86,7 +86,7 @@ beginModule( 'physX3' );
    addIncludePath( $PHYSX3_SDK_PATH . "/Include/common" );
       
    // Libs
-   addProjectLibDir( $PHYSX3_SDK_PATH . "/Lib/win32" );
+   addProjectLibDir( $PHYSX3_SDK_PATH . "/Lib/vc10win32" );
    addProjectLibInput( "PhysX3_x86.lib","PhysX3CHECKED_x86.lib" );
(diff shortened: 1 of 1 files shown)


eightyeight:
@rextimmy could this be implemented in the Project Manager UI for modules?

rextimmy:
I think it could if you pre-choose the visual studio version you where using in the UI. Or maybe if the project gen itself had some functions to determine what VS version it is currently building.

JeffProgrammer:
@rextimmy @eightyeight CHECKED should be changed to DEBUG for debug builds in project generator for the DLL/lib files, since CHECKED extension lib has errors with torque on 3.3.2

rextimmy:
Yes definitely, i should have done this some time ago. It was only originally included because CHECKED libs have support for PVD. You will get better performance using CHECKED libs in release mode anyway if you really do need to check things in the PVD. It's just impossible to cover every scenario with the project manager.

eightyeight:
?

rextimmy:
Sorry i was cleaning out old branches and didn't realise this was still open

eightyeight:
:+1: It'd be great if we could figure out some way to fix this up for various VS versions. Could we have several different modules for different VS versions, which you can select one from with the UI? Similar to the way the move manager selection option works? Then we could have several small modules with the relevant config settings that differ.

PR #999, opened 2014-12-01
Fix so it compiles correctly on Visual Studio 2013

Winterleaf:
This fixes a bug when compiling on visual studio 2013


LuisAntonRebollo:
Not necesary for compile on VS2013

PR #1000, opened 2014-12-01
FMod switching DLLS if 64 bit.

Winterleaf:


eightyeight:
Do we need a linux equivalent? This looks good though.

Areloch:
Looks good to me.

PR #1001, opened 2014-12-01
Bringing some parameters up to speed.

Winterleaf:
Update to the max number of ghosts, as well as to default connection speeds.

The connection speeds reflect the high speed nature the internet is now.

### DIFF
--- a/Engine/source/console/sim.h
+++ b/Engine/source/console/sim.h
@@ -64,7 +64,7 @@ typedef U32 SimObjectId;
 enum SimObjectsConstants
 {
    DataBlockObjectIdFirst = 3,
-   DataBlockObjectIdBitSize = 10,
+   DataBlockObjectIdBitSize = 14,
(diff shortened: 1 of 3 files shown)


Areloch:
I'll re-do this PR with the 508 for the packet size as suggested by @dottools 

@dottools do you think that 32 is good on the packet rate, or is leaving it at 10 ideal?

eightyeight:
Sounds good.

dottools:
@Areloch I think it's a safe bet to keep the `gPacketRateToClient` default at 10 Hz and let users/developers change that as they see fit for their own projects.

Areloch:
Sounds good to me. I'll get an updated version of this PR hammered out tomorrow then.

Areloch:
Submitted an updated PR with the suggested paramters: PR #1375 As such, I'll close this PR.

I would still like to thank Winterleaf for looking into updating these in the first place.

PR #1006, opened 2014-12-03
Fix lighting errors when all lights are disabled.

LuisAntonRebollo:
When scene has zero lights sources actives, last ```RenderTexTargetBinManager```render target  are used for render. This cause render artifatcs.

screenshot 2014-12-03 23 48 13

Thx @Lopuska for the fix :)

For replicate:

* Remove or hidde all light sources on editor.


eightyeight:
Um, I don't think @LuisAntonRebollo made another change :P.

eightyeight:
Ok, looks good now :)

PR #1007, opened 2014-12-04
Added scriptable move triggers for AIPlayers. Fixes an issue where AIPla...

GuyAllard:
These changes fix the issues described in [this forum thread](http://www.garagegames.com/community/forums/viewthread/140646).

Fixing the above issue has resulted in the ability to set any supported move trigger on an aiplayer from script. For example, to make bob crouch:
bob.setMoveTrigger($player::crouchtrigger);


eightyeight:
Something seems to have broken the merge. I'll do it manually today.

PR #1027, opened 2014-12-10
Allow normals on shadowed surfaces

bpay:
Changed ambient for use a factor of light with camera direction, this allow normals on shadowed surfaces.

Before: 
screenshot 2014-12-08 20 54 47

After:
screenshot 2014-12-08 20 54 58

(on behalf of Luis)


LuisAntonRebollo:
We tested at BeamNG and improve a lot visuals.

I'm going to merge, we can revise later for a better implementation,

PR #1048, opened 2014-12-13
missing texture format.

Lopuska:
To silence a console error on startup (OpenGL)


Azaezel:
need to review how opengl handles dds B8G8R8A8s. (seems to swizzle the channels)

LuisAntonRebollo:
About BGRA8 discuss #1163

PR #1049, opened 2014-12-13
Add VS2012 support to projectgenerator

cornytrace:
Files are essentially just copied from the 2010 version with slight
tweaks, works without any warnings or errors.
It prevents having to do the annoying conversion from the VS2010 version every time you regenerate.
Is additional work on projectgenerator even wanted, seeing as there is also the competing cmake build system?


LuisAntonRebollo:
@cornytrace , have you signed OSA? 

We need your GG account name for verification.

LuisAntonRebollo:
I get this error: ``` RC : fatal error RC1110: could not open Torque.rc ```

I think we need to copy some files to ```templates/???/buildFiles/VisualStudio 2012/projects```

PR #1080, opened 2014-12-26
Use a strong reference instead of more manual reference counting

eightyeight:
Replaces part of #1078. Instead of doing manual reference counting, we use a `StrongRefPtr` in this method.


LuisAntonRebollo:
Thx :)

Issue #774, opened 2014-09-05
TorqueScript anonymous functions


eightyeight:
As in [this resource](http://www.garagegames.com/community/resources/view/16723) and popular other languages. Anonymous functions are great. However, we should probably decide on an identifier for them which is not `function`, so as to [not treat every function as a lambda](https://github.com/JeffProgrammer/BeyondReality/commit/f492e44b0362eef583f2920dad0aaa2a094e0c49), but also because JavaScript has shown us that typing out `function` everywhere is annoying. Additional enhancement would be to allow anonymous function bodies to have no curly braces around them, so we can write very nice short function bodies.

I propose using a simple backslash as the lambda introducer; to whit:

    // Using a traditional block;
    schedule(1000, 0, \(%name) {
       echo("Hello" SPC %name);
    }, "Peter");

Additionally, we should support calling functions from variables:

    %sch = "schedule";
    %sch(1000, MyObject, myMethod);

    %fn = \(%x) %x + 1;
    %y = %fn(5);

PR #1090, opened 2014-12-29
Anonymous functions

eightyeight:
Slight modification of [Konrad's resource](http://www.garagegames.com/community/resource/view/16723) for #774. Example:

    schedule(2000, 0, function(%a) { echo(%a); }, horses);


eightyeight:
For some reason the number of lines in the generated parser seems to have dropped significantly. Hrm. Ah well. Let's do it!

PR #1095, opened 2015-01-05
Simple check for NULL fix. The crash was happening when trying to create...

antonyjones67:
... box collision for a shape in the Shape Editor. I am not sure if this is the proper fix because because I would think the list would shrink from 3 to 2 instead of just setting the last element to NULL. Didn't spend a lot of time with this to find out what the larger problem was.


dottools:
Needs to change tabs to 3 spaces per each tab character and it should be fine. I do find it strange that there are NULL entries to begin with though.

antonyjones67:
yeah. this is a band-aid. Lots of instability working with collision in 
the shape editor. who is the right person for this? should i pursue?

dragonsoulz:
Trying to update?

PR #1108, opened 2015-01-18
Modified files for SDL2.

LuisAntonRebollo:
Engine modifications for allow new SDL platform layer.

### DIFF
--- a/Engine/source/gui/editor/guiMenuBar.h
+++ b/Engine/source/gui/editor/guiMenuBar.h
@@ -32,6 +32,7 @@
 
 class GuiMenuBar;
 class GuiMenuTextListCtrl;
+class WindowInputGenerator;
 
 class GuiMenuBackgroundCtrl : public GuiControl
 {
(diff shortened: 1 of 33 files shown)


Engine/source/gui/editor/guiMenuBar.h
```diff
@@ -161,21 +163,25 @@ class GuiMenuBar : public GuiTickCtrl //  Was: GuiControl
 
 	// internal menu handling functions
 	// these are used by the script manipulation functions to add/remove/change menu items
-
+   static Menu* sCreateMenu(const char *menuText, U32 menuId);
```
dottools: Why is this not `createMenu`? I assume `s` short for static, of which doesn't follow the rest of the static method naming conventions.

eightyeight: True, static members are usually prefixed but methods not so much.

LuisAntonRebollo: Ok, i can fix this after merge for avoid remake the PRs.

Thx for review :)

PR #1114, opened 2015-01-19
fix a freeze when you repeatly open-close the gui editor pressing F10

Lopuska:


Azaezel:
also fixes the hit x when gui editor is open bug.

eightyeight:
Weird. Doesn't the loop properly delete all elements in the menu bar?

bpay:
@eightyeight same question here. Doesn't this just introduce a memory leak? Why does this work?

LuisAntonRebollo:
I think it's better dont change code.

```MenuBar``` and ```PopupMenu``` are not derived from SimGroup and dont delete childs at destruction

**Comment by eightyeight** [comment]
So we need to figure out where the crash is coming from :P. Maybe try `> 0` instead of `!= 0`?

**Comment by Lopuska** [comment]
Now we this changes it works. What do you think? :)

**Comment by eightyeight** [comment]
Won't that leave one item (index `0`) in the menu bar?

**Comment by Lopuska** [comment]
changing "i > 0" to "i >= 0" causes again the freeze :(
Maybe @LuisAntonRebollo  has a better Idea for fixing it.

**Comment by fishulla** [comment]
maybe, getCount returns the number of elements not the index so the range would be 1 to n and not 0 to n.

**Comment by Lopuska** [comment]
When the index "i" is equal to 0:
tools/guiEditor/scripts/guiEditorCanvas.ed.cs (223): Unable to find object: '' attempting to call function 'delete'
Set::getObject - index out of range.
It's out of range, so it should be correct the current solution.

**Comment by dottools** [comment]
This fix is still wrong. Truth table testing:
## Test Data
Assume the following occurred before the follow up code examples are called:
```PHP
%this.menuBar.add(%menu1);
%this.menuBar.add(%menu2);
```
Two menus will exist in this truth table test.

### Old Code
This code is simple enough. Just keep deleting the first element in the container until it's empty.
```PHP
 while( %this.menuBar.getCount() > 0 )
 %this.menuBar.getObject( 0 ).delete();
```
%this.menuBar.getCount() | %this.menuBar.getObject( 0 )
------------ | -------------
2 | _Valid Object_
1 | _Valid Object_
0 | _Not Called_


### Current Code
I assume this code is meant to destroy the objects in reverse order they were created like a stack pop, but it's off by one.
```PHP
 for( %i = %this.menuBar.getCount(); i > 0; %i-- )
 %this.menuBar.getObject( i ).delete();
```
%this.menuBar.getCount() | %i | %this.menuBar.getObject( %i )
------------ | ------------- | -------------
2 | 2 | **Out of Bounds**
2 | 1 | _Valid Object_
1 | 0 | _Not Called_


### Suggested Code
Small change now makes this routine function properly.
```PHP
 for( %i = %this.menuBar.getCount(); i > 0; %i-- )
 %this.menuBar.getObject(i -1).delete();
```
%this.menuBar.getCount() | %i | %this.menuBar.getObject(%i -1)
------------ | ------------- | -------------
2 | 2 | _Valid Object_
1 | 1 | _Valid Object_
0 | 0 | _Not Called_

Now it successfully removes all menus in the menu bar successfully. However this doesn't answer as to why the current code changes fixes the issue by leaving a menu entry in the menu bar while correctly destroying all menu entries in the menu bar is causing the issue. Further testing and debugging still needs to be done to determine what is really causing the lag problem.

Lopuska:
That's cool dottools :)
Thank you for providing this great review.

dottools:
Funny enough after posting my comment I just realized this code change is completely invalid since i instead of %i is used in condition expression field `i > 0` vs `%i > 0`.

It should be this:
```PHP
 for( %i = %this.menuBar.getCount(); %i > 0; %i-- )
 %this.menuBar.getObject(%i -1).delete();
```

So who knows what the code was doing without `%` sigil...

Lopuska:
Thanks. updated with the correct version.

eightyeight:
> So who knows what the code was doing without `%` sigil...

Comparing a string literal `i` to 0, which I suspect will always return `false`, as `i` will parse as `0`.

Areloch:
As far as I can tell, this has largely stopped being an issue, likely from some other changes that were rolled in?

I cannot reproduce, and with the SDL menubar fix i did, I think that nips the menu crash issue in the butt.

If no one else can replicate it with the current devhead, I'd consider this issue closed.

Areloch:
Going to close this for now as it seems to have largely been resolved via other, more root changes to the guis and menubars and the like.

If the issue crops back up, we can revisit this fix.

PR #1116, opened 2015-01-19
Intel OpenGL - fix a crash/shader compile error when you activate Vertex...

Lopuska:
... Lit from Material Editor


LuisAntonRebollo:
Confirmed :+1:

PR #1118, opened 2015-01-20
fix #1117

### LINKED ISSUE #1117: OpenGL bug - removing diffuse from a skybox cause shadergen error
The problem is caused because 'attn' for some reason is a float (scalar) when we don't have diffuse.
so during acess of -> 'float.xxxx' we get an error.

I fixed that changing this:   

```cpp
  {
      if ( attn )
         lerpVal = new GenOp( "saturate( @ ).xxxx", attn );
      else
         blendOp = Material::Mul;
   }
```
to:

```cpp
   {
      if ( attn && fd.materialFeatures.hasFeature( MFT_DiffuseMap ) )
         lerpVal = new GenOp( "saturate( @ ).xxxx", attn );
      else
         blendOp = Material::Mul;
   }
```

Lopuska:
fix for #1117 bug


LuisAntonRebollo:
Confirmed the problem and fix.

Thx :+1:

PR #1119, opened 2015-01-20
OpenGL fix - fixed a crash when you activate Alpha Threshold checkbox wi...

Lopuska:
...thout a diffuse map


LuisAntonRebollo:
Confirm problem and solution is good... thx :)

For easy reproduce:

* Use debug build
* Use Opengl render
* Enter mission
* Open world editor
* Change to material editor
* You have BlankWhite as selected
* Remove Diffuse texture
* Activate alpha threshold

PR #1121, opened 2015-01-21
Fix buffer overflows

bpay:
Fixes about 15 actual buffer overflows due to incorrect use of sizeof on UTF16 strings, and provides safer functions to use for static sized buffers. We found that some of these (in fileDialog.cpp) were regularly writing into caller's stack frames.


Azaezel:
verified

bpay:
Rebased against current development

LuisAntonRebollo:
Great work :)

:+1:

PR #1124, opened 2015-01-25
Forest wind emitter

eightyeight:
Redo of #896 with cleanup.

### DIFF
--- a/Engine/source/forest/forestWindEmitter.cpp
+++ b/Engine/source/forest/forestWindEmitter.cpp
@@ -39,6 +39,8 @@
 #include "T3D/gameBase/processList.h"
 #include "console/engineAPI.h"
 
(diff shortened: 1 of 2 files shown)


Areloch:
Works fine for me.

eightyeight:
What does the effect actually look like? I haven't see it in action, just refactored Vince's code. How can you see the sphere if it intersects the camera location?

Areloch:
It pretty much forces it to render as much as it can. You still see some clipping  - I think due to issues with the near plane - but you can at least tell the circle still there and didn't inexplicably stop existing.

There's probably an even better way to do it for the future, but this should work well enough as a fix for the current method.

eightyeight:
Double-sided sphere?

Issue #1127, opened 2015-01-25
OpenGL Basic Lighting rendering issues


JeffProgrammer:
![](http://i.imgur.com/i66OfDc.jpg)

ew. I suspect something with glowing?

JeffProgrammer:
I replaced the road texture with one from @Azaezel and all the glowing problems went away with both road and water...

JeffProgrammer:
i need to check to see if current fixes in opengl have fixed this issue...

JeffProgrammer:
Still happening.
![](http://i.imgur.com/EBuLCPj.png)

JeffProgrammer:
OKAY, SO GET THIS ONE FOLKS.

I ran cleanShaders.bat and it goes away....renders fine in BL on GL:

![](http://i.imgur.com/e9qlePr.jpg)

eightyeight:
If by fine you mean 'it's night on the land and day at sea' :joy:

LuisAntonRebollo:
@JeffProgrammer , can you compare x64 and x86 builds?

Here i get the problem in x64 only :(

LuisAntonRebollo:
I have fixed a problem with basic water and HDR.

@JeffProgrammer, can you retest?

PR #1129, opened 2015-01-26
revised road top texture

Azaezel:
texture mentioned in https://github.com/GarageGames/Torque3D/issues/1127


Areloch:
I don't mind replacing it with a nicer texture, but do we have any idea why changing the road texture resolved his water glowing?

eightyeight:
Yep, looks like some underlying issue :P.

Issue #1125, opened 2015-01-25
Check OpenGL peformance on Linux/Windows


LuisAntonRebollo:
...

JeffProgrammer:
I will give you the results on OSX once I get it working.

JeffProgrammer:
On windows in outpost, I'm seeing a 6.66mspf decrease in OpenGL using a GTX 750ti with latest drivers.  Using latest dev branch as of 1/25/2015 @ 6:30pm EST

JeffProgrammer:
@LuisAntonRebollo I'm actually seeing a ~9-10mspf increase in openGL now. :( Release mode, latest drivers, with the get* thing merged.

97fps D3D, 55fps Ogl

JeffProgrammer:
scatter sky is the culprit!

It's taking 5mspf extra to render on my system. replace that in outpost with a sun and you shall see.

Azaezel:
Some back and forth brainstorming and tracing on this one seems to indicate issues with scattersky in particular, as the source of a 6ms hit GL side.

LuisAntonRebollo:
OK, thx :)

When 3.7 are a bit more stable i need to do a opengl performance pass.

PR #1130, opened 2015-01-26
Remove a get* OpenGL function causing CPU-GPU sync point.

LuisAntonRebollo:
Fix bad performance on OpenGL.

Related with #1125


eightyeight:
:+1:

Issue #1136, opened 2015-01-28
Bullet build is broken by linux changes


dragutux:
removing stdint.h fix the problems but basically https://github.com/GarageGames/Torque3D/commit/4e9034854d057f745f7449502cbad03a3fe8b76e#diff-637773112ccf1788179bc28c358075e0R26 create a redefinition of uint64 for bullet!

dragutux:
PR coming with a proper fix (using a define from bullet)

LuisAntonRebollo:
Thx for report this problem :)

```stdint.h``` is a standart library supported on VS2010, the only problem is we need to configure Bullet library.

I think use T3D header entry point for Bullet is better to place the fix, for avoid add library code to a general header.

dragutux:
err where exactly ?

LuisAntonRebollo:
Fixed on #1146

LuisAntonRebollo:
Fixed and merged on Fix Bullet compilation. #1198

PR #1138, opened 2015-01-28
commit to fix issue in https://github.com/GarageGames/Torque3D/issues/1136

dragutux:
this fix the redefinition of uint64_t by stdint.h (in the case of bullet, it does not like it, you either have to remove stdint... or disable their specific call)


LuisAntonRebollo:
i think we can do a clean fix... see #1136 coment.

Lopuska:
That is the same job I've done before my porting of GMK.
Confirm is functional, this PR is a really important step for bullet and future GMK port.
Thanks for this PR.

PR #1139, opened 2015-01-28
Cmake unity build

bpay:


eightyeight:
Very cool. I'll give this a try when I can. It seems like a great idea, though!

jamesu:
Might be worth investigating if you can use precompiled headers which might be a more pragmatic solution to build times.

bpay:
Precompiled headers aren't a good fit for Torque for a several reasons:
1. They are best suited to builds where you are using lots of external headers that almost never change. In Torque, we are mostly including our own headers, which change often since they frequently contain inline code. For this reason I doubt we would see much win from PCH.
2. They aren't platform independent. Each compiler has a different implementation of PCH, of varying degrees of quality, and CMake doesn't provide any high-level abstraction for these. So there would be the additional burden of maintaining and testing the PCH systems on all the different platforms, whereas the unity build is the same everywhere.
3. My hunch is that the speedups you can get with PCH are strictly less than that of unity builds, even for projects where PCH is well suited. In game development where rapid iteration is very important, this is a big consideration.
4. Unity builds essentially get us link time code generation for free, since the compiler is able to inline across cpp files. This is a nice side benefit that you don't get with PCH.
5. PCH reduces compiler parallelism, since the PCH needs to be compiled serially before any other source files. This is a minor problem since it only affects from-scratch builds, but still something to think about.

Overall, I think this is the most pragmatic solution available, in the sense that the pain to gain ratio is minimal. We have recently started using this setup downstream, and I already feel like my productivity is way up, and my frustration with the compiler is way down :) Increasing build speeds by 5x materially changes the way you develop code, for the better (train yourself to use debug builds for the best results). And if you still want the tea-break compiles, you can always switch the unity build off :)

Another plus is that our turnaround time for CI builds is significantly reduced. We are doing the unity build, plus a compile-only build of the non-unity configuration to catch any missing includes. And we do both of those in less time than it took for the single build before. There haven't been any inadvertent breakages from people forgetting includes so far, but I will have more data on this as we continue to use the build.

PR #1141, opened 2015-01-29
Changed type to NetSocket

MusicMonkey5555:
Changed some S32 types to use the typedef NetSocket since they are a
socket.


eightyeight:
Cherry-picked in bc31d79.

PR #1142, opened 2015-01-29
Functions to get EXE Info

MusicMonkey5555:
Added console functions to get info from an executable like version
info, timestamp and more using a library.

### DIFF
--- a/Engine/lib/versinfo/VERSINFO.cpp
+++ b/Engine/lib/versinfo/VERSINFO.cpp
@@ -1,0 +1,145 @@
+//=============================================================================
+//  General component library for WIN32
+//  Copyright (C) 2000, UAB BBDSoft ( http://www.bbdsoft.com/ )
+//
+// This material is provided "as is", with absolutely no warranty expressed
+// or implied. Any use is at your own risk.
+//
+// Permission to use or copy this software for any purpose is hereby granted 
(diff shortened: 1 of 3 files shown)


dottools:
Not sure why this belongs in a game engine and especially since this is Windows only specific. Only Windows executable and dynamic library files can have version info to query while all other platforms do not. This kind of code belongs in a platform specific tool outside of the engine.

MusicMonkey5555:
Makes sense, I sometimes forget about cross platform. The reason we personally use it is to show version numbers and other information inside the game so the user knows what version they have. But I can understand why it would not belong in the head build.

Azaezel:
if that's the inent, may try hooking into https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/source/torqueConfig.h#L37-L43

eightyeight:
https://github.com/GarageGames/Torque3D/blob/development/Engine/source/app/version.h

Areloch:
So, should this just be closed since it'd make more sense to access either the torqueConfig or the version.h as per az and dan's suggestions?

I definitely think there should be script hooks, but probably using one of the above methods instead to keep things playing nice regardless of platform ;)

Lopuska:
Closing this as we prefer wait a more portable/cross-platform solution.
Thank you very much @MusicMonkey5555 as always for your awesome contributions!

PR #1146, opened 2015-01-31
Clean CMAKE modules

LuisAntonRebollo:


LuisAntonRebollo:
Update with  #1148

bpay:
Since you asked for feedback...brace yourself :) I think this is a very good cleanup overall.

1. I am nervous about globbing the .cmake files. The glob order is not well defined, and if there are dependencies between them, there might be breakages that only happen on certain platforms. I can see why people would want to just drop in their own modules, but I think we should put these in a separate directory ("modules/extra" or something), and just list all the built-in modules explicitly.

2. Re: the issue about requiring multiple passes. To avoid this, I think it would be best to just define all the top level options in torque3d.cmake. This way we avoid the problem of dependencies between the options in different modules (such as between TORQUE_ADVANCED_LIGHTING and TORQUE_OPENGL). If there are sub-options that are specific to certain modules, such as TORQUE_SFX_DirectX in module_win32, these can be added cleanly inside the module file with [CMakeDependentOption](http://www.cmake.org/cmake/help/v3.0/module/CMakeDependentOption.html)

3. I think module_core.cmake should just be inlined into torque3d.cmake. It's not *really* a module since it's not optional. Same with module_win32.target.cmake. But I don't feel that strongly about this either way.

Summing up, I think torque3d.cmake would look something like this:

```cmake
# put in basics..
macro(adv_option)
    option(${ARGN})
    mark_as_advanced(${ARGV0})
endmacro()

# options in alphabet order
adv_option(TORQUE_ADVANCED_LIGHTING "Advanced Lighting" ON)
adv_option(TORQUE_BASIC_LIGHTING "Basic Lighting" ON)
#...
#
# modules in alphabet order (unless other order required)
include(modules/module_advanced_lighting.cmake)
include(modules/module_basic_lighting.cmake)
include(modules/module_bullet.cmake)
include(modules/module_extended_move.cmake)
include(modules/module_hifi.cmake)
#....
# inlined module_core.cmake
finishExecutable()
# inlined module_win32_target.cmake
```

eightyeight:
Do you guys have any opinions on how we could add a layer above CMake? Would this affect how these files are designed? For example, I imagine that at some point, the Project Manager might generate a project-specific CMakeLists.txt file based on a per-project configuration (like the current Project Generator config files), which then includes the necessary modules. Am I way off?

galek:
If you want,I can share modules written by me:

1)Fmod Ex(not studio)
2)PhysX3

LuisAntonRebollo:
@galek, yes... we need cmake modules for this :)

Can you add in a new PR?

galek:
@LuisAntonRebollo 
My repository is private and on "bitbucket",so it's impossible.

Modules is needed fix for copying dll to target dir

PhysX3
http://pastebin.com/QTuKzYE1
Fmod
http://pastebin.com/E5d7ULvy

LuisAntonRebollo:
Fit more 3.8 Milestone.

Areloch:
Given that we're going to look at a new project manager with CMake as the backend for 3.9, it makes more sense to push this back so we can give it a bit more attention and be 100% happy with the module configuration(as there still seems to be a bit of design debate going on here) so it can be smoothly utilized by the new project manager.

So, for now, moving this to 3.9

Issue #1144, opened 2015-01-30
Issues on x64 builds


LuisAntonRebollo:
### OSX problems
- [x] @jamesu report: https://github.com/GarageGames/Torque3D/commit/4e9034854d057f745f7449502cbad03a3fe8b76e#commitcomment-9516570 See  #1149
- [x] @jamesu report: https://github.com/GarageGames/Torque3D/commit/fcf7bee64af8dd7b07696bd0bcb99db8dc56b451#commitcomment-9516353 See  #1149
### Changes for new types
- [x] Create a new T3D typedef for `uintptr_t` and `intptr_t` See  #1149
- [ ] Rethink T3D using a ptr to store `SimObjectId` https://github.com/GarageGames/Torque3D/commit/4e9034854d057f745f7449502cbad03a3fe8b76e#commitcomment-9517497

JeffProgrammer:
some carbon things might have to be taken out. we should really replace the whole platformMac from being carbon based to cocoa based as that's both i386 and x86_64

PR #1149, opened 2015-01-31
More x64 fixes

LuisAntonRebollo:
See  Issues on x64 builds #1144

### DIFF
--- a/Engine/source/T3D/cameraSpline.cpp
+++ b/Engine/source/T3D/cameraSpline.cpp
@@ -193,7 +193,7 @@ void CameraSpline::renderTimeMap()
    void *ptr = vb.lock();
    if(!ptr) return;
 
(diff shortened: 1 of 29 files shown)


bpay:
These type aliases look a lot like macros. Can they be lowercased? I know the rest of the numeric typedefs are not, but I think it's too late to fix those...plus, they have numbers embedded (S32, F32) in them which makes them a little less macro-like.

eightyeight:
Typedefs are type-level macros ;).

LuisAntonRebollo:
I only added SPTR and UPTR types, becouse some ppl commented to me that c++ types dont look consistent with T3D types.

I prefer to use c++ style types, but we cant change old types. 

I'm going to separate fix and rename to diferent PRs.

LuisAntonRebollo:
Continue discussion on #1162

PR #1153, opened 2015-02-03
Asserts cleanup

bpay:
Fixes for assert macros and unused variable warnings in release builds


Azaezel:
seems to check out, though for this one I'll note I'm troubled by suppressing warnings...

eightyeight:
I'm fine with that because it only occurs in a release build, and you should always run through debug builds first. However it should probably be wrapped so it only happens in MSVC or we'll get unknown pragma warnings elsewhere, AFAICT.

Lopuska:
I close this as conflicts must be resolved to proceed with merge.
New PR: #1529

Issue #396, opened 2013-05-28
Using LF (\n) instead of CRLF (\r\n) causes some parsing errors.


lukaspj:
Ran a tool on my code which, among other things, replaced all file endings with LF.
This resulted in a lot of parsing errors in some of the script (not all, just some of them 8 in total)

Example of an error:
```
core/scripts/client/audio.cs Line: 414 - syntax error
>>> Advanced script error report.  Line 414.
>>> Some error context, with ## on sides of error halt:
///                     string to pause sources on all channels.
/// @param %pauseSet    An optional SimSet which is filled with the paused 
///                     sources.  If not specified the global SfxSourceGroup 
///                     is used.
///
/// @deprecated
///
function sfxPause( %channels, %pauseSet )
{
## ##  // Did we get a set to populate?
   if ( !isObject( %pauseSet ) )
	  %pauseSet = SFXPausedSet;
	  
   %count = SFXSourceSet.getCount();
   for ( %i = 0; %i < %count; %i++ )
   {
	  %source = SFXSourceSet.getObject( %i );

	  %channel = sfxGroupToOldChannel( %source.getGroup() );
>>> Error report complete.
```
Complete list of files that gave an error:
* core/scripts/client/audio.cs Line: 414 - syntax error
* tools/editorClasses/scripts/projects/projectEvents.ed.cs Line: 27 - syntax error
* tools/editorClasses/scripts/projects/projectInternalInterface.ed.cs Line: 131 - syntax error
* tools/editorClasses/scripts/input/inputEvents.ed.cs Line: 27 - syntax error
* tools/base/menuBar/menuBuilder.ed.cs Line: 193 - syntax error
* tools/worldEditor/scripts/editors/terrainEditor.ed.cs Line: 36 - syntax error
* tools/worldEditor/scripts/editors/terrainEditor.ed.cs Line: 36 - syntax error
* tools/datablockEditor/datablockEditor.cs Line: 338 - syntax error

Windows8x64 MIT 3.0 with no changes except line endings.

JeffProgrammer:
Looking into this...

JeffProgrammer:
the regular expression requires at least 1 character after the 3 slash marks in order for it to compile from what I see, the regular expression:

[code]("///"[^/][^\n\r]*[\n\r]*)+ { return(Sc_ScanDocBlock()); }[/code]

PR #1156, opened 2015-02-06
Fix issue 396

JeffProgrammer:
This fixes issue #396 

The regex expression assumed that the programmer would type a character after the beginning of the summary comment line. This now stops it from requiring an extra char.

This will now compile:
```
///
function pq()
{
}
```


eightyeight:
I believe the meaning is slightly different now. Previously `////` was not parsed ad a doc block, because the grammar specified `[^/]` immediately after `"///"`. Now the not-slash is optional, and in the case of four slashes, the fourth slash will be parsed as part of the `[^\n\r]` block.

Reflecting, I think the solution should actually be `("///"([^/][^\n\r]*)?[\n\r]*)+`. Will that work?

JeffProgrammer:
someone slap me to do this, keep forgetting.

eightyeight:
:hand:

JeffProgrammer:
should be fixed now. Can someone confirm?

eightyeight:
Lol CI is totally broken. Hurrah. Works for me in linux.

JeffProgrammer:
glad to hear its working good on linux too ;)

eightyeight:
Shoot. Apparently it doesn't actually fix the issue on Windows. Wat.

eightyeight:
[This](https://github.com/eightyeight/Torque3D/commit/2c6cced0cf97f057a2aa3a8e2352efb96c47f0df) was the best I could do.

JeffProgrammer:
my fix compiles scripts bytecode fine though on windows, What do you mean?

eightyeight:
Did you test all scripts? With `\r\n` replaced by `\n`? For me, your fix didn't work on `audio.cs` with newlines replaced. I actually found a way to keep current behaviour w.r.t `////`. See #1220.

PR #1157, opened 2015-02-07
Show the canvas immediately in unix because the splash doesn't work

eightyeight:
`displaySplashScreen` doesn't work, meaning nothing appears for ages. So on unix we'll show the canvas immediately even though it looks a little ugly.


eightyeight:
Build failure apparently caused by failure to fetch repo...?

PR #1158, opened 2015-02-07
Specify rpath=. when compiling in linux

eightyeight:
This allows you to move the executable and not get shared-object errors.

### DIFF
--- a/Tools/CMake/torque3d.cmake
+++ b/Tools/CMake/torque3d.cmake
@@ -30,8 +30,8 @@ if(UNIX)
     
     # default compiler flags
     # force compile 32 bit
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXX_FLAG32} -Wall -Wundef -msse -pipe -Wfatal-errors ${TORQUE_ADDITIONAL_LINKER_FLAGS}")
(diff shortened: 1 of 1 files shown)


bpay:
Are you intending to set the RPATH to the current directory of the executable? If so, the way to do this is to specify $ORIGIN. Otherwise you are setting the rpath to the current working directory which is a no-no from a security standpoint.

see: https://groups.google.com/d/msg/gnu.gcc.help/G3E68okFJHs/J_HGBNsajuYJ

eightyeight:
Hmm, okay. Using . worked for me when setting it through cmake flags on the command line. I'll try again with origin. Any idea why CI isn't accepting this? maybe this'll just have to always be a command-line thing people enable for themselves.

eightyeight:
Going to add this to documentation instead. Looks like it'll be a pain to make all the various configurations handle this sensibly.

eightyeight:
And hilariously, I get this error when trying to use `$ORIGIN` on the command line:

    CMake Error at Engine/lib/sdl/CMakeLists.txt:1244 (add_library):
      Target "SDL2" links to item " -Wl,-rpath=$ORIGIN" which has leading or
      trailing whitespace.  This is now an error according to policy CMP0004.

The command was

    cmake ../../../.. -DTORQUE_APP_NAME=Empty -DTORQUE_ADDITIONAL_LINKER_FLAGS="-Wl,-rpath,\$ORIGIN"

Though I think actually this is just an error, and can be ignored - I seem to be able to build with it. It also looks like the problem is somewhere in SDL's CMake files, not ours :pensive:

PR #1159, opened 2015-02-07
Include navigation and testing modules by default

eightyeight:
* They are both useful and can easily be turned off.
 * Navigation being off by default causes problems with the navmesh editor, which is _included_ by default, but nonfunctional unless the module is included.
 * Navigation can kind of be considered a stock feature.

The only problem is that I see no way to do something equivalent with the project generator. We could add `includeModule` to the Templates' default config files, but this won't show up in the manager UI when you create a new project.

### DIFF
--- a/Engine/source/T3D/aiPlayer.cpp
+++ b/Engine/source/T3D/aiPlayer.cpp
@@ -28,6 +28,8 @@
 #include "T3D/gameBase/moveManager.h"
 #include "console/engineAPI.h"
 
+#include <cfloat>
+
 static U32 sAIPlayerLoSMask = TerrainObjectType | StaticShapeObjectType | StaticObjectType;
 
 IMPLEMENT_CO_NETOBJECT_V1(AIPlayer);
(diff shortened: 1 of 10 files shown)


LuisAntonRebollo:
Fixed compilation on Linux.

eightyeight:
T_T thanks for that, I meant to come back and check it but then bedtime happened. Also it appears there's no way to do this in the project manager. I think. I will try to make that happen.

eightyeight:
Project manager 2.2 will allow setting default modules for inclusion, so I think this is good to go.

eightyeight:
Current error:

    ../../../../Engine/source/platform/test/platformTimerTest.cpp:68:54: error: no matching function
    for call to ‘JournaledSignal<void(int)>::notify(TimeManager_BasicAPI_Test::TestBody()::handle*,
    void (TimeManager_BasicAPI_Test::TestBody()::handle::*)(S32))’
        time.timeEvent.notify(&handler, &handle::timeEvent);
                                                          ^

:sweat:

PR #1164, opened 2015-02-07
Remove some dead code from OpenGL shadergen.

LuisAntonRebollo:


JeffProgrammer:
All is good. confirm it works. Also remove my continue; statement in the PR you merged with my change, not needed now :) +1

LuisAntonRebollo:
Thx for test.

I'm going to preserve continue, if posible in future we need to extend this code.

Is correct and safe :)

JeffProgrammer:
sounds good :)

Issue #1134, opened 2015-01-28
Pull Request #1026 broke the console stack


jamesu:
There have been numerous reports that the console stack is throwing out asserts in the development branch. I've tracked it down to a certain pull request of mine, PR 1026 (https://github.com/GarageGames/Torque3D/pull/1026). Intended to fix issue #976 .

It seems while it fixed the reported issue, it has the unintended side-effect of not cleaning the stack out correctly during future function calls.

jamesu:
I have a preliminary fix for this which resolves this by changing the stack behaviour of executef and the callback helpers (i.e. it makes them all go through templates which properly manage the stack). I'd like to write some basic tests however to ensure the new behaviour isn't breaking anything...

eightyeight:
Is there a consistent way to repro this? At the moment my way of testing it will just involve playing around with the engine to make sure nothing looks broken :P. Are there any unit tests among those you've added that fail before, and pass after the PR?

jamesu:
There are unit tests now for the execute stack handling.

ATM Most stuff works with the new code, the problem I guess is some minor things handle console stack values strangely which don't have tests and thus need to manually be tested...

eightyeight:
Yeah, such as that `dFree` bug earlier. This is why we want to have a pre-release :joy:

jamesu:
It looks like there is a problem with the string iterators. Looks a bit messy, will see if I can work around it.

jamesu:
Should be noted that using dIsSpace is a pretty bad idea here as on some platforms the behavior will change depending on locale settings.

jamesu:
Looks like startsWith is somehow corrupting the value in the string buffer. VERY strange.

jamesu:
Upon further investigation it seems the console stack is getting resized WHILE the string iterator is being used, causing invalid pointers. Gonna have to wrap up that pointer with a helper class so it is offset from the correct position.

jamesu:
Pushed a fix for this which uses a relative pointer type. In addition I've added a new console variable type which acts as a pointer to a stack string which should help reduce the overall stack size (by reusing the data in the previous frames of the stack).

eightyeight:
![](http://upload.wikimedia.org/wikipedia/commons/4/47/River_terrapin.jpg)

Will test soon.

PR #1165, opened 2015-02-08
Fix for console stack

jamesu:
Refactors console execution in order to resolve issue #1134. It does this by changing executef calls to templates which internally manage the stack, similar to how most modern C++ script binding libraries work. The templates also automatically perform type conversion, meaning you don't even need to bother converting your parameters to TorqueScript types - this is all done for you.

Along with this, all public execution calls (such as execute* and evaluate*) now save and restore the stack upon returning. They are also now thread safe, meaning they can be called from other threads (as long as get*Arg functions aren't used).

I've also added some basic tests for the core marshalling functionality in addition to the execution calls. This should hopefully trap any expected future regressions with this functionality.

In addition some callbacks have been corrected to use the correct types instead of strings which was causing a few problems.

### DIFF
--- a/Engine/source/console/console.h
+++ b/Engine/source/console/console.h
@@ -44,6 +44,8 @@ struct ConsoleFunctionHeader;
 class EngineEnumTable;
 typedef EngineEnumTable EnumTable;
 
+typedef U32 StringStackPtr;
+
 template< typename T > S32 TYPEID();
 
 
@@ -121,8 +123,9 @@ public:
    
    enum
    {
-      TypeInternalInt = -4,
-      TypeInternalFloat = -3,
+      TypeInternalInt = -5,
+      TypeInternalFloat = -4,
+      TypeInternalStringStackPtr = -3,
       TypeInternalStackString = -2,
       TypeInternalString = -1,
    };
@@ -166,6 +169,7 @@ public:
    S32 getSignedIntValue();
    F32 getFloatValue();
    const char *getStringValue();
+   StringStackPtr getStringStackPtr();
    bool getBoolValue();
    
    void setIntValue(U32 val);
@@ -173,6 +177,7 @@ public:
    void setFloatValue(F32 val);
    void setStringValue(const char *value);
    void setStackStringValue(const char *value);
+   void setStringStackPtrValue(StringStackPtr ptr);
    void setBoolValue(bool val);
    
    void init()
@@ -187,7 +192,7 @@ public:
    void cleanup()
    {
       if (type <= TypeInternalString &&
-          sval != typeValueEmpty && type != TypeInternalStackString )
+          sval != typeValueEmpty && type != TypeInternalStackString && type != TypeInternalStringStackPtr)
          dFree(sval);
       sval = typeValueEmpty;
       type = ConsoleValue::TypeInternalString;
@@ -219,6 +224,7 @@ public:
    static ConsoleValueRef fromValue(ConsoleValue *value) { ConsoleValueRef ref; ref.value = value; return ref; }
 
    const char *getStringValue() { return value ? value->getStringValue() : ""; }
+   StringStackPtr getStringStackPtrValue() { return value ? value->getStringStackPtr() : 0; }
 
    inline U32 getIntValue() { return value ? value->getIntValue() : 0; }
    inline S32 getSignedIntValue() { return value ? value->getSignedIntValue() : 0; }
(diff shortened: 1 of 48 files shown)


eightyeight:
Ninja build fails:

    ../../../../Engine/source/console/console.h:877:122: fatal error: use 'template' keyword to treat 'call' as a dependent template name
            template<typename A> ConsoleValueRef executef(A a) { _EngineConsoleExecCallbackHelper<A> callback( a ); return callback.call<ConsoleValueRef>(); }
                                                                                                                                    ^
                                                                                                                                    template 

[This](http://stackoverflow.com/questions/3786360/confusing-template-error) may shed some light.

jamesu:
Added a workaround. Hopefully should resolve the problem your end.

eightyeight:
CI appears jammed waiting for executors on the master machine. Not sure what's up.

jamesu:
Looks like for some reason there's a problem with the include paths on Linux. Will have to check it out tomorrow.

Azaezel:
afraid I've got some bad news on this one. After applying https://github.com/GarageGames/Torque3D/pull/1171 to correct the query master method, applying this fix-attempt seems to have exposed a leak. breaks 'randomly' within https://github.com/GarageGames/Torque3D/blob/development/Engine/source/platform/platformMemory.cpp#L1750-L1791
latest stacktrace: http://pastebin.com/nc20Ds6H

jamesu:
@azaezel do you have any reproduction steps? A stack trace isn't too useful here as I need to see what's happening leading up to said leak

Azaezel:
simplest repro case once that others corrected is just hit query master server

jamesu:
Might have to rearrange these headers. GCC seems to handle these templates pretty badly.

jamesu:
Fixed GCC build. Having a look to see whats going on with your issue @Azaezel

eightyeight:
Thanks :). I'll see if I can reproduce sometime soon as well.

jamesu:
Ok mystery solved. Looks like someone forgot to remove a couple of dFree calls ( see https://github.com/GarageGames/Torque3D/blob/59a5261b0298cdd04f1c7ba684e8d85b9dff2057/Engine/source/app/net/serverQuery.cpp#L436 ) which ended up freeing two of the string values in the console stack.

eightyeight:
Whoops, that was introduced in #1072. Glad it was someone else's fault though.

Engine/source/console/console.h
```diff
@@ -872,21 +847,47 @@ namespace Con
...
+	template<typename A, typename B, typename C, typename D, typename E, typename F, typename G, typename H, typename I> ConsoleValueRef executef(A a, B b, C c, D d, E e, F f, G g, H h, I i) { _EngineConsoleExecCallbackHelper<A> callback( a ); return callback.template call<ConsoleValueRef>(b, c, d, e, f, g, h, i); }
+	template<typename A, typename B, typename C, typename D, typename E, typename F, typename G, typename H, typename I, typename J> ConsoleValueRef executef(A a, B b, C c, D d, E e, F f, G g, H h, I i, J j) { _EngineConsoleExecCallbackHelper<A> callback( a ); return callback.template call<ConsoleValueRef>(b, c, d, e, f, g, h, i, j); }
+	template<typename A, typename B, typename C, typename D, typename E, typename F, typename G, typename H, typename I, typename J, typename K> ConsoleValueRef executef(A a, B b, C c, D d, E e, F f, G g, H h, I i, J j, K k) { _EngineConsoleExecCallbackHelper<A> callback( a ); return callback.template call<ConsoleValueRef>(b, c, d, e, f, g, h, i, j, k); }
+	/// }
```
Azaezel: going to need to add 
   template<typename A, typename B, typename C, typename D, typename E, typename F, typename G, typename H, typename I, typename J, typename K, typename L> ConsoleValueRef executef(A a, B b, C c, D d, E e, F f, G g, H h, I i, J j, K k, L l) { _EngineConsoleExecCallbackHelper<A> callback(a); return callback.template call<ConsoleValueRef>(b, c, d, e, f, g, h, i, j, k, l); }
Because reasons.

eightyeight:
Variadic templates FTW...

eightyeight:
Merged this into devhead, clean&rebuild, run debug and hit 'GUI editor' and I get this:

error

jamesu:
Will have a look

eightyeight:
Info I should have provided - assert fired in a call to `dIsSpace` with some random input string. The char val was `-18`, and the assert was checking for the range `0-255`.

jamesu:
Repost from the issue page...

Pushed a fix for this which uses a relative pointer type. In addition I've added a new console variable type which acts as a pointer to a stack string which should help reduce the overall stack size (by reusing the data in the previous frames of the stack).

eightyeight:
Hilariously, CI appears to be fine, but when I try to compile this locally I get 

    /torque/Engine/source/console/console.cpp:1760:33: error: cast from ‘char*’ to ‘U32 {aka unsigned int}’ loses precision [-fpermissive]
           return STR.mBuffer + (U32)sval;
                                     ^
    compilation terminated due to -Wfatal-errors.

Any suggestion as to what I should do to that line? Or why some builds might be fine with it?

jamesu:
Looks like your doing a 64bit build, in which case it would be better to use UINTPTR. Will push a fix later.

eightyeight:
Ah, that makes sense. Maybe CI isn't doing x64 at the moment. @LuisAntonRebollo?

eightyeight:
It's working for me in Windows x32, though I never saw the issues this was supposed to fix. Good to go, pending x64 compatibility :).

eightyeight:
Oh my gosh CMake is freaking irritating. Every time I change a single source file and `make`, here we go, try to build every dependency project (lucky they're all already built), scan the engine project for dependencies, recompile a whole slew of source files, and _then_ see what the error is now.

eightyeight:
@jamesu I recommend, if you have time, following [this](http://wiki.torque3d.org/coder:compiling-using-vagrant) guide to compile the engine in Ubuntu x64. It would tighten the development loop, rather than me finding errors I don't understand, and you then fixing them blindly! I'm going to have another go fixing things on this end to recommend some changes.

eightyeight:
See 57eed9c for the changes I had to make to placate GCC.

Issue #1166, opened 2015-02-08
Gestalt no longer works on Mac OSX Yosemite


JeffProgrammer:
The Gestalt() function was deprecated since Mac OSX Mountain Lion. It no longer works properly on Yosemite, as the operating system returns mavericks while running Yosemite. A replacement will have to be made for newer versions of OSX.

(Not surprising it stopped working, it was a really old Carbon API function...)

I know It's kinda early to start reporting OSX bugs, but I am doing some work on OSX after all... ;)

JeffProgrammer:
I got this. /me self assigns.

eightyeight:
Thanks for the report and fix!

PR #1173, opened 2015-02-09
remove gestalt() as it has been deprecated and stopped working on Mac OS...

JeffProgrammer:
... Yosemite

Fixes Issue #1166


JeffProgrammer:
This does work fine on my mac. Do we need someone not using the latest OS to confirm that this works OK? The docs do state that it does work on 10.2+

PR #1175, opened 2015-02-10
feedback for *which* namespace is already linked. 

Azaezel:
should probably look into enhancing assertwarn to be able to take a formatted string at some point...


jamesu:
I just use `avar("%s", ....)`

PR #1178, opened 2015-02-10
DiffuseMapFeatGLSL::setTexData validation

Azaezel:
this shouldn't ever trigger, but for paranoias sake, best check it.


Lopuska:
That's was already checked in the past, we have removed this IF statement with Luis to resolve a bug!
Check on commit history for more info.

Azaezel:
after a consult, as unintuative as it sounds, may need to go the opposite of this, and strip out the if (tex) checks for the rest. will dig about a bit more, before offering that as an alternative...

eightyeight:
Heh. This is why I don't like to merge things quickly :joy:

Issue #1179, opened 2015-02-10
setMathControlState is not implemented on Mac, Linux, or X64 Windows


jamesu:
Platform::setMathControlState does not appear to be implemented for Mac, Linux, or X64 Windows. This function turns out to be very important as otherwise you will get weird rounding errors on floating point operations during execution.

jamesu:
Still not implemented it for x64 windows, though you should be able to use the same asm. Unless there is something blocking this?

eightyeight:
Nothing blocking :).

LuisAntonRebollo:
@jamesu, we really need to handle this FPU changes?

We use a flag for avoid DX9 change FPU precission, and new DX or OpenGL dont change precission.

jamesu:
@LuisAntonRebollo yes, since the state of the FPU is not guaranteed across platforms (AFAIK). For instance on linux we had big problems with the precision of a float in an AI calculation causing a crash. Best to be explicit IMO ;)

LuisAntonRebollo:
Ok,thx for info :)

Re-opened for future research.

PR #1180, opened 2015-02-10
Add math control state functions for intel

jamesu:
Fixes issue #1179


eightyeight:
Should this be wrapped in a `define` detecting the CPU type?

jamesu:
For osx, no.

For windows, there is a compiler function available though that Its not supported by gcc afaik. Nobody uses winrt. So I guess no.

For linux, yes though I am unaware of any non-x86 linux platforms which t3d officially supports.

PR #1182, opened 2015-02-11
Bug fixes

MusicMonkey5555:
After working with torque for a while I have noticed several bugs. Those bugs have been fixed in this branch.

### DIFF
--- a/Engine/source/gui/controls/guiConsoleTextCtrl.cpp
+++ b/Engine/source/gui/controls/guiConsoleTextCtrl.cpp
@@ -115,6 +115,9 @@ void GuiConsoleTextCtrl::onPreRender()
    {
       mResult = Con::evaluatef( "$guiConsoleTextCtrlTemp = %s;", mConsoleExpression.c_str() );
       
+      //Fixes a bug with the above not always grabbing the console text.
+      mResult = Con::getVariable("$guiConsoleTextCtrlTemp");
+      
       // Of the resulting string we will be printing,
       // Find the number of lines and length of each.      
       mProfile->mFont->wrapString( mResult, U32_MAX, mStartLineOffset, mLineLen );
--- a/Engine/source/gui/controls/guiConsoleTextCtrl.cpp
+++ b/Engine/source/gui/controls/guiConsoleTextCtrl.cpp
@@ -113,7 +113,7 @@ void GuiConsoleTextCtrl::onPreRender()
 {   
    if ( mConsoleExpression.isNotEmpty() )
    {
-      mResult = Con::evaluatef( "$guiConsoleTextCtrlTemp = %s;", mConsoleExpression.c_str() );
+      Con::evaluatef( "$guiConsoleTextCtrlTemp = %s;", mConsoleExpression.c_str() );
       
       //Fixes a bug with the above not always grabbing the console text.
       mResult = Con::getVariable("$guiConsoleTextCtrlTemp");
--- a/Engine/source/gui/controls/guiMLTextCtrl.cpp
+++ b/Engine/source/gui/controls/guiMLTextCtrl.cpp
@@ -168,7 +168,7 @@ DefineEngineMethod( GuiMLTextCtrl, scrollToTag, void, (S32 tagID),,
    object->scrollToTag( tagID );
 }
 
-DefineEngineMethod( GuiMLTextCtrl, scrollToTop, void, ( S32 param1, S32 param2),,
+DefineEngineMethod( GuiMLTextCtrl, scrollToTop, void, (),,
    "@brief Scroll to the top of the text.\n\n"
    "@tsexample\n"
    "// Inform GuiMLTextCtrl object to scroll to its top\n"
@@ -631,7 +631,7 @@ bool GuiMLTextCtrl::setCursorPosition(const S32 newPosition)
       mCursorPosition = 0;
       return true;
    }
-   else if (newPosition >= mTextBuffer.length()) 
+   else if (newPosition >= mTextBuffer.length() - 1)
    {
       mCursorPosition = mTextBuffer.length();
       return true;
@@ -669,11 +669,11 @@ void GuiMLTextCtrl::getCursorPositionAndColor(Point2I &cursorTop, Point2I &curso
 {
    S32 x = 0;
    S32 y = 0;
-   S32 height = mProfile->mFont->getHeight();
+   S32 height = mProfile && mProfile->mFont ? mProfile->mFont->getHeight() : 0;
    color = mProfile->mCursorColor;
    for(Line *walk = mLineList; walk; walk = walk->next)
    {
-      if ((mCursorPosition <= walk->textStart + walk->len) || (walk->next == NULL))
+      if ((mCursorPosition < walk->textStart + walk->len) || (walk->next == NULL))
       {
          // it's in the atoms on this line...
          y = walk->y;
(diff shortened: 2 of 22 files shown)


Engine/source/gui/controls/guiMLTextCtrl.cpp
```diff
@@ -669,11 +669,11 @@ void GuiMLTextCtrl::getCursorPositionAndColor(Point2I &cursorTop, Point2I &curso
 {
    S32 x = 0;
    S32 y = 0;
-   S32 height = mProfile->mFont->getHeight();
+   S32 height = mProfile && mProfile->mFont ? mProfile->mFont->getHeight() : 0;
```
dottools: This needs to be the following as parenthesis is very important to avoid turning `height` into a true/false evaluation result:
```C++
S32 height = (mProfile && mProfile->mFont) ? mProfile->mFont->getHeight() : 0;
```

Engine/source/gui/controls/guiMLTextCtrl.cpp
```diff
@@ -768,7 +768,7 @@ void GuiMLTextCtrl::onMouseDown(const GuiEvent& event)
    mSelectionAnchorDropped = event.mousePoint;
    if (mSelectionAnchor < 0)
       mSelectionAnchor = 0;
-   else if (mSelectionAnchor >= mTextBuffer.length())
+   else if (mSelectionAnchor >= mTextBuffer.length() - 1)
```
dottools: This is wrong, it's suppose to be:
```C++
   else if (mSelectionAnchor >= mTextBuffer.length())
      mSelectionAnchor = mTextBuffer.length() -1;
```

Engine/source/math/mPoint3.h
```diff
@@ -855,7 +859,8 @@ inline F64 Point3D::lenSquared() const
 
 inline F64 Point3D::len() const
 {
-   return mSqrtD(x*x + y*y + z*z);
+	F64 temp = x*x + y*y + z*z;
```
dottools: A released optimized build would fix this for us, but it really isn't necessary to duplicate expressions when you can help it. Also tabs vs 3 spaces issue.
```C++
   F64 temp = x*x + y*y + z*z;
   return (temp > 0.0) ? mSqrtD(temp) : 0.0;
```

Engine/source/math/mQuat.cpp
```diff
@@ -32,33 +32,48 @@ const QuatF QuatF::Identity(0.0f,0.0f,0.0f,1.0f);
...
+	sysz = sy*sz;
+	sycz = sy*cz;
+	cysz = cy*sz;
+	w = cycz*cx + sysz*sx;
+	x = cycz*sx + sysz*cx;
+	y = sycz*cx - cysz*sx;
+	z = cysz*cx - sycz*sx;
+	*/
+	// Assuming the angles are in radians.
+	F32 c1 = mCos(e.y/2.0f);
```
dottools: Concerned about changing from multiply to division code considering floating point division is the most expensive operation you could do on any hardware.
Why aren't these lines this?:
```C++
   F32 c1 = mCos(e.y * 0.5f);
   F32 s1 = mSin(e.y * 0.5f);
   F32 c2 = mCos(e.z * 0.5f);
   F32 s2 = mSin(e.z * 0.5f);
   F32 c3 = mCos(e.x * 0.5f);
   F32 s3 = mSin(e.x * 0.5f);
```

Engine/source/scene/sceneContainer.cpp
```diff
@@ -1012,6 +1012,11 @@ bool SceneContainer::_castRay( U32 type, const Point3F& start, const Point3F& en
       F32 currStartX = normalStart.x;
 
       AssertFatal(currStartX != normalEnd.x, "This is going to cause problems in SceneContainer::castRay");
+	  if(_isnan(currStartX))
```
dottools: Oops tab vs 3 spaces issue.

eightyeight:
Thanks for all these. Nathan's pointed out a lot of things that should be fixed. I'll review in more detail after 3.7 most likely. Build breaks on Ubuntu:

    ../../../../Engine/source/scene/sceneContainer.cpp:1015:7: fatal error: use of undeclared identifier '_isnan'
          if(_isnan(currStartX))
             ^

Engine/source/scene/sceneContainer.cpp
eightyeight: And apparently `_isnan` is not included on Linux.

Engine/source/math/mQuat.cpp
MusicMonkey5555: Yeah that would be more optimized :) I just noticed that the conversion wasn't correct and got the code from someplace and it worked. Think a friend.

Engine/source/math/mPoint3.h
MusicMonkey5555: For sure!

Engine/source/scene/sceneContainer.cpp
MusicMonkey5555: hmm @eightyeight is there some alternative? or could some wrapper code be added? Would love for torque to be multi-platform. For our projects we have just used windows, so I often don't take the time to make sure it's multi-platform code we write.

I suppose a platform check could be added, but this was added because of a bug we experienced.

MusicMonkey5555: hmm here is a post about it: http://stackoverflow.com/questions/570669/checking-if-a-double-or-float-is-nan-in-c

Engine/source/gui/controls/guiMLTextCtrl.cpp
MusicMonkey5555: I can't remember what issue I was having with this, but I remember it causing some bug. When reading the code it makes sense that it should be what it was. But something was happening. I will take a look and talk with other people that worked with our code.

MusicMonkey5555: Oh ok so it looks like your change should have the same behavior, there was no -1 for mSelectionAnchor assignment. I actually noticed we made several little -1 changes to selection stuff as the selection had odd behavior the specifics of each change was back before we use version control and I think dates back to TGE so can't dig those up, but since this change should have the same result then that works. I will make the change.

Engine/source/scene/sceneContainer.cpp
MusicMonkey5555: Guess there is mIsNaN_F in mMathFn.h so I will change it to use that.

Engine/source/gui/controls/guiConsoleTextCtrl.cpp
```diff
@@ -115,6 +115,9 @@ void GuiConsoleTextCtrl::onPreRender()
    {
       mResult = Con::evaluatef( "$guiConsoleTextCtrlTemp = %s;", mConsoleExpression.c_str() );
```
eightyeight: Please remove the `mResult =` here, it makes the control flow confusing to have two assignments to the same variable.

eightyeight:
Thanks for all this! Looking forward to getting it tested and rolled in.

Azaezel:
seems to check out

eightyeight:
Pending response from @MusicMonkey5555

MusicMonkey5555:
Sorry for not being very active. Our company closed due to restructuring of Boeing and so I have been actively looking for a job, working on personal projects and taking care of a 1+ year old. So I've had hardly any time to keep up with my commits.

Amazing how not having a job can keep you busier than having one :)

eightyeight:
Oh gosh, sorry to hear that! I imagine it can. I hope things work out!

PR #1183, opened 2015-02-11
Documentation

MusicMonkey5555:
I added/fixed some documentation. If it was incorrect documentation it was changed, if there were inconsistencies in variable names I changed them and if something wasn't documented I added it.

### DIFF
--- a/Engine/source/gui/controls/guiTreeViewCtrl.cpp
+++ b/Engine/source/gui/controls/guiTreeViewCtrl.cpp
@@ -4872,7 +4872,7 @@ DefineEngineMethod( GuiTreeViewCtrl, removeSelection, void, ( S32 itemId), ,
    "Deselect an item or remove it from the selection.\n\n"
    "@param itemId Item Id to deselect.\n")
 {
-	object->removeSelection(itemId);
+   object->removeSelection(itemId);
 }
 
 DefineEngineMethod( GuiTreeViewCtrl, removeChildSelectionByValue, void, ( S32 parentId, const char* value), ,
@@ -5012,7 +5012,7 @@ DefineEngineMethod( GuiTreeViewCtrl, getItemText, const char *, ( S32 itemId ), 
    "@param itemId TreeItemID to get text of.\n"
    "@return Text for a given item.")
 {
-	return(object->getItemText(itemId));
+   return(object->getItemText(itemId));
 }
 
 DefineEngineMethod( GuiTreeViewCtrl, getItemValue, const char *, ( S32 itemId ), ,
@@ -5049,7 +5049,7 @@ DefineEngineMethod( GuiTreeViewCtrl, removeAllChildren, void, (S32 itemId), ,
 DefineEngineMethod( GuiTreeViewCtrl, clear, void, (), ,
    "Empty the tree.\n")
 {
-	object->removeItem(0);
+   object->removeItem(0);
 }
 
 DefineEngineMethod( GuiTreeViewCtrl, getFirstRootItem, S32, (), ,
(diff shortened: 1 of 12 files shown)


Engine/source/gui/worldEditor/worldEditor.h
```diff
@@ -92,7 +92,6 @@ class WorldEditor : public EditTSCtrl
       S32 getSelectionSize();
       S32 getSelectObject(S32 index);	
       const Point3F& getSelectionCentroid();
-      const char* getSelectionCentroidText();
```
eightyeight: What happened here?

Engine/source/gui/worldEditor/worldEditor.cpp
```diff
@@ -3449,8 +3500,18 @@ DefineConsoleMethod( WorldEditor, transformSelection, void,
...
+   "@param point Position to transform by."
+   "@param relativePos True to use relative position."
+   "@param rotate True to transform the selection's rotation."
+   "@param rotation Rotation to transform by."
+   "@param relativeRot True to use the relative rotation."
+   "@param rotLocal True to use the local rotation."
+   "@param scaleType Scale type to use."
+   "@param scale Scale to transform by."
+   "@param sRelative True to use a relative scale."
+   "@param sLocal True to use a local scale.")
```
eightyeight: Oh gosh, thanks for adding something to this monster...

eightyeight:
Great stuff :). Should be easy to merge.

Engine/source/gui/controls/guiTreeViewCtrl.cpp
```diff
@@ -4858,24 +4858,33 @@ DefineEngineMethod( GuiTreeViewCtrl, addSelection, void, ( S32 id, bool isLastSe
...
    object->addSelection(child->getID());
 }
 
-DefineConsoleMethod(GuiTreeViewCtrl, removeSelection, void, (S32 id), , "(deselects an item)")
+DefineEngineMethod( GuiTreeViewCtrl, removeSelection, void, ( S32 itemId), ,
+   "Deselect an item or remove it from the selection.\n\n"
+   "@param itemId Item Id to deselect.\n")
 {
-   object->removeSelection(id);
+	object->removeSelection(itemId);
```
eightyeight: Tab

Engine/source/gui/controls/guiTreeViewCtrl.cpp
```diff
@@ -4952,95 +4989,135 @@ DefineConsoleMethod( GuiTreeViewCtrl, setItemImages, void, ( S32 id, S8 normalIm
...
 }
 
-DefineConsoleMethod(GuiTreeViewCtrl, getItemText, const char *, (S32 index), , "(TreeItemId item)")
+DefineEngineMethod( GuiTreeViewCtrl, getItemText, const char *, ( S32 itemId ), ,
+   "Gets the text for a given item.\n\n"
+   "@param itemId TreeItemID to get text of.\n"
+   "@return Text for a given item.")
 {
-   return object->getItemText(index);
+	return(object->getItemText(itemId));
```
eightyeight: Tab

Engine/source/gui/worldEditor/worldEditor.h
MusicMonkey5555: So if I recall correctly the only reason this was here is to output to a console method, after converting it over to use a return console type of Point3F it isn't really needed. Guess this isn't so much documentation, but in the act of converting over console functions to use the new way of defining them, it was a necessary change. Hopefully that all makes sense. Just look at line 3252 or 3261 in the .cpp file.

Engine/source/gui/controls/guiTreeViewCtrl.cpp
MusicMonkey5555: Yeah where I work we settled on tabs instead of spaces, so making any merges with GG stuff is a pain. Luckily VS has the format feature. I do apologizes though for them!

eightyeight:
The tabs issue is tricky since so many files use tabs already, inconsistently :P. Look, I'm going to go ahead and merge this, even if it increases the tab-noise a little. At some point in the future we'll make some sort of style pass over the engine.

eightyeight:
Weird that this passed CI. It appears to break my local compile thanks to `scrollVisible` now returning `bool`. Will fix it.

PR #1184, opened 2015-02-11
JSON support

MusicMonkey5555:
Added a JSON class that is open to script  and inherits from the array object. It is basically the same as the array object only with some added functions/variables to handle the type handling and writing out to a string.

There is no reading in, perhaps that can be added by someone else or in the future.

### DIFF
--- a/Engine/source/console/jsonObject.cpp
+++ b/Engine/source/console/jsonObject.cpp
@@ -75,84 +75,84 @@ void JSONObject::initPersistFields()
 
 const JSONObject::DataType JSONObject::getValueType( const String& val ) const
 {
-	JSONObject* jsonObj = NULL;
-	if(val.isEmpty())
-	{
-		return DataType::UNDEFINED;
-	}
-	else if (!val.equal("0", String::NoCase) && Sim::findObject(val, jsonObj))
-	{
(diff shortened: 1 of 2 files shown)


Lopuska:
That's really useful for multiplayer and network related gameplay like leaderboard and social media integration that expose a json api.
I'll take a look this soon! 
Thank you for contribution.

dottools:
Pretty sure TAML already has JSON support. Could just use that instead.

Lopuska:
@dottools Well, I'm gonna to check this. If that's the case, It's better to avoid this PR for redundant feature and more useless code to maintain.
Thanks for that hint.

Lopuska:
https://github.com/GarageGames/Torque3D/tree/development/Engine/source/persistence/rapidjson
https://github.com/GarageGames/Torque3D/tree/development/Engine/source/persistence/taml/json

I close this as redundant feature. No reason to have duplicate json parser class.
Probably @MusicMonkey5555 implemented this when TAML was still not present in T3D.
TAML commits have been pushed to repo 4 months ago as this PR is much older.
Thanks @dottools for your good eye :P

Anyway, thank you very much @MusicMonkey5555 to share your great work with us.
Next days i'm going to see your other pending PRs.

PR #1186, opened 2015-02-11
Tcp additions

MusicMonkey5555:
Added several little additions to the TCPObject such as sending a file, callbacks for when everything is finished, and extra script methods.

### DIFF
--- a/Engine/source/app/net/tcpObject.cpp
+++ b/Engine/source/app/net/tcpObject.cpp
@@ -407,21 +407,21 @@ void TCPObject::send(const U8 *buffer, U32 len)
 
 bool TCPObject::sendFile(const char* fileName)
 {
-	//Open the file for reading
-	FileStream readFile;
-	if(!readFile.open(fileName, Torque::FS::File::Read))
-	{
-		return false;
-	}
+   //Open the file for reading
+   FileStream readFile;
+   if(!readFile.open(fileName, Torque::FS::File::Read))
+   {
+      return false;
+   }
 
-	//Read each byte into our buffer
-	Vector<U8> buffer(readFile.getStreamSize());
-	readFile.read(buffer.size(), &buffer);
+   //Read each byte into our buffer
+   Vector<U8> buffer(readFile.getStreamSize());
+   readFile.read(buffer.size(), &buffer);
 
-	//Send the buffer
-	send(buffer.address(), buffer.size());
+   //Send the buffer
+   send(buffer.address(), buffer.size());
 
-	return true;
+   	return true;
 }
 
 DefineEngineMethod(TCPObject, send, void, (const char *data),, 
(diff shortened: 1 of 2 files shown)


Engine/source/app/net/tcpObject.cpp
```diff
@@ -400,6 +405,29 @@ void TCPObject::send(const U8 *buffer, U32 len)
...
+	//Read each byte into our buffer
+	Vector<U8> buffer(readFile.getStreamSize());
+	U8 byte;
+	while(readFile.read(&byte))
+	{
+		buffer.push_back(byte);
+	}
+
+	//Send the buffer
+	send(buffer.address(), buffer.size());
```
jamesu: This is not guaranteed to send all the data, especially with a large file. What you need is a mechanism to send chunks of the file data per tick.

jamesu: Also `while(readFile.read(&byte))` and pushing each byte to the vector has to be the worst way I can imagine of reading a file. Use `read(numBytes, buffer)` at least. :)

eightyeight: I remembered you submitted something similar to this James, but didn't have time to check if it was exactly this.

jamesu: This was https://github.com/GarageGames/Torque3D/pull/658 - however I pulled it due to there not being "a real consensus on this about the implementation". From what I can tell the only similarity between @MusicMonkey5555 's changes (in total) and that branch are the incorporation of the bytesWritten parameter in Net::send.

MusicMonkey5555: Interesting about the file size. I guess we never hit that limit, but make sense. I wrote a bunch of script code that basically used TCP object to do a bunch of HTTP stuff including opening up a SocketStream to send stuff to and from Google Earth. Most of the files we were sending were pretty small art files or like a JS file so there wasn't an issue. I will look into it and try to improve it, but it works in those limited cases.

And I know there is an HTTP object, but we needed to make rapid changes and the HTTP object didn't seem to:
1. do everything we need
2. had some issues with it working at all

As for the performance issue with reading the file, I'll improve that with your suggestion, thanks!

MusicMonkey5555: Oh also yeah the bytesWritten part I actually got from you code. My comment on running into an issue when it was queued was on there and I just send the entire file as it worked for what we needed it for. So they are somewhat related. My comments on it still stand though that you run into issues with that implementation of a send queue.

MusicMonkey5555:
After taking a look at the latest changes to the file I noticed all the WebSocket reading support wasn't added, so you could only write. Therefore I added that to this in case someone wants it.

Lopuska:
Good additions, also a very secure change as doesn't modify anything, just adds! Particularly important for network code. Thank you.

PR #1187, opened 2015-02-11
Added functions

MusicMonkey5555:
This is a plethora of added functions, improvements to existing functionality, opening of some things to script and other misc changes that everyone could benefit from.

I was going to try to break it up into more branches, but most of these things are nice and it would be a lot of pull request. Also I am not sure how I would group them. I am willing to remove things or change code that people thing aren't.


dottools:
First glace there's some promising improvements, however tabs vs 3 spaces issues too. ;)

eightyeight:
Yeah, tabs everywhere. Not sure how we'd deal with that... rebasing the lot would be a massive pain. Appreciate the effort splitting out all the commits though! :)

Engine/source/T3D/shapeBase.cpp
```diff
@@ -3600,6 +3637,38 @@ DefineEngineMethod( ShapeBase, playThread, bool, ( S32 slot, const char* name ),
...
+
+   "@return The max number of threads (0 to ShapeBase::MaxScriptThreads-1).\n\n"
+   
+   "@see pauseThread()\n"
+   "@see stopThread()\n"
+   "@see setThreadDir()\n"
+   "@see setThreadTimeScale()\n"
+   "@see destroyThread()\n")
+{
+   return ShapeBase::MaxScriptThreads;
```
jamesu: Seems a bit odd having a function which returns a constant which could just be set as a console variable

Engine/source/console/consoleObject.h
```diff
@@ -542,6 +545,19 @@ class AbstractClassRep : public ConsoleBaseType
 
 extern AbstractClassRep::FieldList sg_tempFieldList;
 
+inline AbstractClassRep* AbstractClassRep::getClassRepFromName(const char* className)
+{
+	AbstractClassRep* current = classLinkList;
+	while(current != NULL)
+	{
+		if(dStricmp(current->getClassName(),className) == 0)
```
jamesu: Since getClassName() returns a StringTableEntry it would be better to convert className to a StringTableEntry then you can just compare pointers instead of using a string compare.

Engine/source/console/simFieldDictionary.cpp
```diff
@@ -168,6 +168,25 @@ SimFieldDictionary::Entry *SimFieldDictionary::findDynamicField( StringTableEntr
...
+   {
+		Entry **walk = &mHashTable[i];
+		Entry *field = NULL;
+		while(*walk)
+		{
+			field = *walk;
+			mVersion++;
+
+			if( field->value )
+				dFree(field->value);
```
jamesu: I would set field->value to NULL too otherwise the entry will be placed back into the smFreeList with an invalid pointer.

Engine/source/console/simObject.cpp
```diff
@@ -2662,17 +2733,108 @@ DefineConsoleMethod( SimObject, getDynamicFieldCount, S32, (),,
...
+{
+	char fieldName[128];
+	char fieldValue[128];
+
+	const char* fn = object->getStaticField(index);
+
+	if(fn)
+		dStrcpy(fieldName, fn);
+	else
+		dStrcpy(fieldName, "");
```
jamesu: if the length of the field name or value is > 128, this will cause a buffer overflow. Similar case below.

Engine/source/console/simSet.cpp
```diff
@@ -1029,6 +1129,50 @@ DefineConsoleMethod( SimSet, getFullCount, S32, (), , "() Get the number of dire
...
+         sum += getHiddenCountRecursive(child);
+   }
+   return sum;
+}
+
+DefineEngineMethod( SimSet, getHiddenCountRecursive, S32, (),,
+   "Get the number of hidden objects contained in the set and all child sets.\n"
+   "@return The number of hidden objects contained in the set and all child sets." )
+{
+   return getHiddenCountRecursive(object);
```
jamesu: Probably nitpicking here, but what possible practical use could this function have?

Engine/source/math/mMatrix.h
```diff
@@ -212,7 +214,23 @@ class MatrixF
    /// Convenience function to allow people to treat this like an array.
    F32& operator ()(S32 row, S32 col) { return m[idx(col,row)]; }
    F32 operator ()(S32 row, S32 col) const { return m[idx(col,row)]; }
-
+   
+   bool operator==(const MatrixF& a) const
+   {
+      return (a[0] == m[0] && a[1] == m[1] && a[2] == m[2] && a[3] == m[3]
```
jamesu: Wouldn't memcmp be a better idea here?

Engine/source/math/mMatrix.h
```diff
@@ -257,6 +275,14 @@ inline MatrixF& MatrixF::set( const EulerF &e, const Point3F& p)
    return (*this);
 }
 
+inline bool MatrixF::equals(const MatrixF& mat)
+{
+   return (m[0] == mat[0] && m[1] == mat[1] && m[2] == mat[2] && m[3] == mat[3] &&
```
jamesu: Same as operator==. Could be called by operator== instead of unnecessarily duplicating.

Engine/source/math/mathTypes.cpp
```diff
@@ -215,6 +225,31 @@ ConsoleSetType( TypePoint3F )
...
+   dSprintf(returnBuffer, bufSize, "%d %d %d %d", pt->x, pt->y, pt->z, pt->w);
+   return returnBuffer;
+}
+
+ConsoleSetType( TypePoint4I )
+{
+   if(argc == 1)
+      dSscanf(argv[0], "%d %d %d %d", &((Point4I *) dptr)->x, &((Point4I *) dptr)->y, &((Point4I *) dptr)->z, &((Point4I *) dptr)->w);
+   else if(argc == 4)
+      *((Point4I *) dptr) = Point4I(dAtoi(argv[0]), dAtoi(argv[1]), dAtoi(argv[2]), dAtoi(argv[3]));
```
jamesu: Should be noted that argc is always 1 here (this is a case for a console feature which was never implemented), thus this code will never be called.

Engine/source/platform/platformNet.cpp
```diff
@@ -127,7 +128,35 @@ static const char* strerror_wsa( S32 code )
       E( WSAEACCES );
 #undef E
       default:
-         return "Unknown";
+		  LPWSTR errString = NULL;  // will be allocated and filled by FormatMessage
+
+		  S32 size = FormatMessageW(  FORMAT_MESSAGE_ALLOCATE_BUFFER |
```
jamesu: Doesn't this return a WIDE string? Likely needs conversion to UTF8.

Engine/source/console/consoleObject.h
eightyeight: It really begs the question, what's the greater cost? A call to `insert` or a string comparison?

jamesu: I'd bet my money on the string comparison since there are tons of classes with varying name lengths in Torque. ;)

Engine/source/T3D/shapeBase.cpp
MusicMonkey5555: True that would probably be a better way to handle it. I will look at changing that.

Engine/source/console/simSet.cpp
MusicMonkey5555: If I remember correctly I was wanting to display the number of hidden objects selected or something.

Engine/source/math/mMatrix.h
MusicMonkey5555: Probably :) Just thought it odd there wasn't a comparison operator. so I added one. I'll make the change.

Engine/source/math/mathTypes.cpp
MusicMonkey5555: Yeah then that should be the case for all the others. I think I just copied and pasted another and changed some stuff. If I make that change we should probably make it for all the types.

Engine/source/platform/platformNet.cpp
MusicMonkey5555: It seemed to work in the one case it came up, but I will look into it.

MusicMonkey5555: It actually takes a LPWSTR and that's why I used one, but not sure if passing that to the getReturnBuffer call will cause issues or not. I would probably need some more input on what the best thing to do would be.

Engine/source/console/consoleObject.h
eightyeight: Oh wait, I should read more carefully. `dStricmp` is called inside a loop. AFAICT the string table only provides any benefit if you have a single `insert` and multiple comparisons. Otherwise the cost of `insert` will outweigh the cost of a comparison :P.

Engine/source/math/mathTypes.cpp
eightyeight: Yup :+1:

Engine/source/math/mMatrix.h
bpay: I wouldn't. memcmp won't give the right answers for floats, e.g. 0.0 wouldn't compare equal to -0.0 as it should. Guessing it would be slower to call memcmp too.

PR #1189, opened 2015-02-11
Added invalid text visual feedback

MusicMonkey5555:
Now instead of just having a sound when you enter bad text there is some visual feedback as well.


dottools:
Conflicts with a disabled control state of which is what ColorNA is used for. Of which NA means Not Available. Which on normal Window UI is the 'grayed out' state control color.

eightyeight:
It's a good idea, but as @dottools says there should be a separate colour for this state.

MusicMonkey5555:
I think I just used that because it wasn't being used in any rendering for that control. I can added it to the guiTypes though and commit that. Suppose it may be a nice feature to add to other controls if need be in the future.

Lopuska:
Thank you! I'm going to review this soon! I close this PR as a duplicate of all-in-one PR: #1334

PR #1196, opened 2015-02-14
return the result value of scrollVisible

brakhane:
The console function changed from void to bool, so we
should actually return the value.

Also, the missing return is an error in VS2013.


Azaezel:
verified necessary for compiliation with the latest devhead.

eightyeight:
Sorry, meant to do this last night then got sidetracked. Thanks :).

PR #1199, opened 2015-02-15
Added NULL check in function findItemByName.

antonyjones67:
(cherry picked from commit adb2513a2461092486e00b5550bf57fb83d0ccac)

When items in the gui tree control are removed it does  not resize the array because it can create a problem with the tree item id's. So it just NULL's out the element. This causes an access violation.


eightyeight:
Thanks for the PR! Testing now. Note - in future, it's a good idea to make a separate branch off `development` for each PR, otherwise there will be pain.

antonyjones67:
sorry i'll do better in the future. still learning this git hub thing :)

Issue #1192, opened 2015-02-13
MiniDump support doesnt get defined when checked off in the project manager


adambeer:
When checking off MiniDump Support in the project manager, #define TORQUE_MINIDUMP doesnt get defined anywhere. This issue is in conjunction with #1191 and both defines are required in order to get minidump support working.

Issue #1191, opened 2015-02-13
#define TORQUE_RELEASE doesnt get defined for release builds


adambeer:
When browsing through the torqueConfig.h, I noticed that TORQUE_RELEASE wasn't defined when it was a release build but TORQUE_DEBUG would define properly.

eightyeight:
I'm pretty sure `TORQUE_RELEASE` is intended to be enabled manually when you actually make a release, not just when you make a release build, maybe? Or possibly I'm thinking of `TORQUE_SHIPPING`.

adambeer:
Yea the manual one is TORQUE_SHIPPING.

PR #1200, opened 2015-02-15
Fix for minidump support

adambeer:
This allows minidump support to compile and be setup correctly when the checkbox is checked. The VS lib templates were missing TORQUE_RELEASE in the preprocessor definitions which caused minidump support to not be enabled regardless of a checkbox or not.

### DIFF
--- a/Engine/source/platformWin32/minidump/winStackWalker.h
+++ b/Engine/source/platformWin32/minidump/winStackWalker.h
@@ -28,6 +28,8 @@
 #include <windows.h>
 #include <dbghelp.h>
 
+#include "platform/types.h"
+
 class StackWalker
 {
 public:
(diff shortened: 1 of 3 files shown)


adambeer:
This is in response to #1191 and #1192

eightyeight:
Do we need to do anything about the `TORQUE_MINIDUMP` define or is that one okay after all?

adambeer:
TORQUE_MINIDUMP is being set correctly, this fix is just for adding TORQUE_RELEASE for release builds which is required along side the MINIDUMP define, and also to get a file to compile when its included from the 2 defines.

eightyeight:
Thanks for the fix!

eightyeight:
@adambeer could you sign the OSA agreement in your GG.com account? I should have checked before merging :).

adambeer:
Done. I knew about this and probably should have done it before submitting ;)

eightyeight:
Thanks!

Issue #1205, opened 2015-02-16
NONINSTANTIABLE_CLASS not showing up in editor.


lukaspj:
If you implement a class with `IMPLEMENT_NONINSTANTIABLE_CLASS` and you use addField TYPEID<>() of the class, then you will get the parent class listed in the editor instead of noninstantiable class.

Not sure if clear enough.. I'm tired :P

lukaspj:
Solved in #1535

PR #1206, opened 2015-02-16
Add support for abstract ConObjects

lukaspj:
Now abstract classes can be properly implemented in a way that makes them work with the editors (I know I'm rambling... Tired.. Ask for a better explanation if necessary :P )

Seeks to solve #1205

### DIFF
--- a/Engine/source/console/consoleObject.h
+++ b/Engine/source/console/consoleObject.h
@@ -553,100 +553,120 @@ extern AbstractClassRep::FieldList sg_tempFieldList;
 /// @see AbtractClassRep
 /// @see ConsoleObject
 template< class T >
-class ConcreteClassRep : public AbstractClassRep
+class ConcreteAbstractClassRep : public AbstractClassRep
 {
-   public:
-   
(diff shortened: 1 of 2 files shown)


lukaspj:
Okay so I'm a little clearer in my head now..

So this really seeked to solve an issue that came up with the ParticleSystemRefactor. 

If you have a structure like this:

* GameBase
  * AbstractClass
    * ConcreteClass

Then the editor will think that GameBase is the immediate parent, and not the AbstractClass. This seeks to fix that.

eightyeight:
Are these script-only namespace/classes? Or actual C++ inheritance? Also when does the editor care about the inheritance hierarchy? The inspector?

Azaezel:
datablock editor dropdowns and the like. Noticed when taking a first-pass review on the particle system refactor, that some conditions could trigger invalid lists. This'll fix that.

lukaspj:
@eightyeight 

Aww :(
First

Yaay!
Second

First one is without this PR, second is with.

Lopuska:
Closing it for conflicts issues New PR: #1535

PR #1212, opened 2015-02-20
Fixed SDL related header includes for linux

pbrand:
When trying to compile the development branch on Fedora 21 (x86_64) I ran into the following error:
- *fatal error: SDL.h: No such file or directory*

This micro contribution fixes the the SDL related header includes, so that the previously mentioned fatal error is resolved. This way the inclusion of SDL related headers is consistent with the the includes of these headers in other classes of this directory as well.

### DIFF
--- a/Engine/source/platformX86UNIX/threads/semaphore.cpp
+++ b/Engine/source/platformX86UNIX/threads/semaphore.cpp
@@ -25,8 +25,8 @@
 // Instead of that mess that was here before, lets use the SDL lib to deal
 // with the semaphores.
 
-#include <SDL.h>
-#include <SDL_thread.h>
+#include <SDL/SDL.h>
+#include <SDL/SDL_thread.h>
 
 struct PlatformSemaphore
 {
--- a/Engine/source/platformX86UNIX/x86UNIXProcessControl.cpp
+++ b/Engine/source/platformX86UNIX/x86UNIXProcessControl.cpp
@@ -31,7 +31,7 @@
 #include <signal.h>
 
 #ifndef TORQUE_DEDICATED
-#include <SDL.h>
+#include <SDL/SDL.h>
 #endif
 
 //-----------------------------------------------------------------------------
--- a/Engine/source/platformX86UNIX/x86UNIXRedbook.cpp
+++ b/Engine/source/platformX86UNIX/x86UNIXRedbook.cpp
@@ -32,7 +32,7 @@
 #include <string.h>
 #endif
 
-#include <SDL.h>
+#include <SDL/SDL.h>
 
 class SDL_CD; // TODO SDL remove
 


eightyeight:
Thanks for the fix! Glad to hear of people working in Linux. Did you manage to get everything working after this?

LuisAntonRebollo:
We need to do some work for a stable linux version... but we are on the way :)

Not sure why other ppl dont have this problem on Ubuntu and VM.

eightyeight:
@LuisAntonRebollo merge if it looks good!

pbrand:
Unfortunately I have not able to get everything working yet.
I copied the `Empty` template to `My Projects` and ran `make debug` in the `Make_Ded` directory.
Everything goes well until it hits:
- *Debug/Empty/app/mainLoop.cpp.o: In function `StandardMainLoop::handleCommandLine(int, char const**)':*

From then on I receive many many `undefined reference to *****` errors.  
I will look into that soon. If you wish to see the errors, I can provide them. Just let me know how you prefer to receive them e.g. in a link to a file, or embedded in a code block :)

eightyeight:
Interesting. So not using CMake? Any luck following instructions similar to [this](http://wiki.torque3d.org/coder:compiling-in-linux)?

pbrand:
Sorry for the late reply. At first I followed the steps of the readme at github so I just used make.
Now that I tried the instructions at your wiki page (thanks for the link, I didn't know there was another guide) it seems to build when using CMake to generate files with the default configuration.   
Afterwards, I am able to run the executable, although I'm not completely sure what I should see.
Am I supposed to get errors like: Missing file: core/parseArgs.cs! ? I don't have a core directory for instance.

eightyeight:
I'm working on updating the readme and github wiki to have the latest information - the previous method is a little old, though it's weird that it doesn't work any more :joy:

pbrand:
I don't know if you are interested in updating the wiki for fedora as well?
If so, I've created a file with the Fedora equivalent of the Ubuntu libraries and how to install them under Fedora, which you can find [here](https://github.com/delftswa2014/Torque3D/blob/compiling-under-fedora/LibrariesFedora.md)

eightyeight:
Awesome! Which version of Fedora do you think is a reasonable compatibility indicator? I've added the instructions to [this page](http://wiki.torque3d.org/coder:compiling-in-linux).

eightyeight:
@pbrand is this change required for you when using CMake to compile? For me, this breaks compiling with `gcc` in Ubuntu. Is it just me that's screwing something up? CI seems to have had no problems.

pbrand:
Regarding your question about which version of Fedora is a reasonable compatibility indicator. I tried to build this project under Fedora 21 (latest stable release at this moment) and Fedora 20 and both were successful.  
I think, although I haven't tested this, that other distributions who use the RPM package manager (e.g. RHEL, CentOS, (open)SUSE) might be able to follow the fedora guide to some extent as well. I quickly  compared the package naming for [openSUSE](https://en.opensuse.org/openSUSE:Package_naming_guidelines#General_Naming) and [Fedora](http://fedoraproject.org/wiki/Packaging:NamingGuidelines#General_Naming) and found them quite similar.

pbrand:
I had problems when using CMake too. However it would be strange to address that to the includes, as some of the other classes already used the `SDL/SDL.h` before my changes.  
I synced my fork with this repository and rebased my branch `linux-build`. After a `ninja-build clean` followed by `ninja-build` the problem was solved for me. Can you please retry to compile it at your machine and notify me if your problem still occurs?

pbrand:
I have a question by the way. I'm not sure whether I should post it at the forum or open a new issue for it, or that it's not relevant at all. That's why I put it here for now. Just let me know if you don't agree with it, then I'll remove this comment or move it to a more suitable place.

I have two Fedora machines at my disposal, one of them uses the proprietary driver from nVidia, which gives me no trouble, the other however, uses the open source radeon drivers (2x HD6850 in crossfire setup) and gives me the following error:  
The requested renderer, D3D9, doesn't seem to be available. Trying the default, OpenGL.
The OpenGL renderer doesn't seem to be available. Trying the GFXNulDevice.
Attempting to create GFX device: GFX Null Device []
Initializing GFXCardProfiler (GFX Null Device Renderer)
   o Chipset : 'NULL Device'
   o Card    : 'GFX Null Device Card'
   o Version : '0'
   o VRAM    : 0 MB
   - Scanning card capabilities...
   - Loading card profiles...
      - No card profile core/profile/GFXNullDeviceRenderer.cs exists
      - No card profile core/profile/GFXNullDeviceRenderer.NULLDevice.cs exists
      - No card profile core/profile/GFXNullDeviceRenderer.NULLDevice.GFXNullDeviceCard.cs exists
      - No card profile core/profile/GFXNullDeviceRenderer.NULLDevice.GFXNullDeviceCard.0.cs exists
Device created, setting adapter and enumerating modes
% Segmentation fault (core dumped)
 
I don't know if this is on purpose, or that this is a bug. Anyway I would like to look into it and see if I can get it to work. Can you give me some directions, as I can't seem to find why this GFX Null Device get's created instead of acquiring the right data for my card. (I think OpenGL must be able to work, as glxinfo and glxgears both seem to give good output). 
Thank you in advance for your attention to this matter :)

eightyeight:
Heh, that's definitely not on purpose - feel free to raise a separate issue about it. Indeed, the OpenGL driver should be selected.

Hmm, I wonder why the `SDL/SDL.h` path was causing problems with just some files. I'll investigate further. It's probably an issue with my setup :sweat:

pbrand:
So even after the rebase you still experience problems?

eightyeight:
Yeah, bizarrely the `#include` in `processControl.cpp` fails. This is using standard `make` and `gcc`. I just reran `cmake`, let's see how it goes...

pbrand:
Okay, make (still) doesn't work for me either, just CMake -> ninja works for me.

eightyeight:
Okay, got it. Strange. My setup has been causing issues a bit recently, in cases where CI is fine. Not sure if it's because our CMake setup is only tested using ninja/clang or what. I'll have a chat to some of the other SC members about it.

pbrand:
By the way, I just noticed something for the 'Build using Qt Creator' section of the wiki.
I had to go to Projects and override the default 'Make: ninja' command by 'ninja-build' for both 'Build Steps' and 'Clean Steps'. I don't know the command for 'ninja' under Ubuntu, but for fedora it is 'ninja-build'.

eightyeight:
That section is somewhat incomplete just FYI, but thanks, I'll add a note about that.

LuisAntonRebollo:
I get errors, compiler can't get SDL headers :(

> This way the inclusion of SDL related headers is consistent with the the includes of these headers in other classes of this directory as well.
It's posible those ```includes``` are dead code we dont use.

I think those changes make T3D to search on OS libraries instead of T3D lib folder.

pbrand:
Yes, it works fine if I compile the code of your development branch with CMAKE.
Sorry, I was not aware that you wished to use libraries of the lib folder (at the time when I submitted this pull request).
Like I mentioned earlier, for this fix I used the build instructions for version 3.6.3, which showed me to install the `sdl-dev` package, that's why I thought I had to use the library of the system.
The `SDL/SDL.h` should indeed point you to the OS libraries at `/usr/include/SDL`. 
I see that this package is not longer required according to the new build instructions at your wiki.
Maybe that's the difference between the old make files and the new CMAKE method?

LuisAntonRebollo:
Old linux instruction are for DEDICATED build.. used only for game servers.

Sorry we are a bit in the middle of the change :P

PR #1213, opened 2015-02-20
Smart vsync

bpay:
Previously we were defaulting to vsync=off, but in recent versions
of windows this is often a bad idea because desktop compositing ("aero")
effectively enforces a vsync in windowed modes, regardless of our
presentation interval. So turning on vsync in those cases is beneficial
because we won't waste CPU effort rendering frames that are just going to
be discarded by the OS anyways, and frame rate displays will more
accurately represent what the user is actually seeing on the monitor.

To solve this, the engine now tracks separate fullscreen and windowed
vsync modes. Fullscreen defaults to vsync=off. Windowed modes default to
vsync=on if desktop composition is detected, and off otherwise. These
settings are stored in the vsyncFullscreen and vsyncWindowed config
variables. The settings are:

0 : vsync disabled
1 : vsync enabled
2 : auto vsync (default)

These variables aren't intended to be changed directly, and they don't
appear in the graphic options. Any changes to the vsync setting via the
graphics options will removed the "auto" behavior, and this can only be
restored by resetting or editing the config by hand.

The old vsync config variable now only acts as an abstraction to minimize the amount of code that needs to know about the two different vsyncs, and it's been renamed to $video::disableVerticalSync so it won't be persisted.

### DIFF
--- a/Engine/source/gfx/gfxDevice.cpp
+++ b/Engine/source/gfx/gfxDevice.cpp
@@ -49,8 +49,12 @@
 
 GFXDevice * GFXDevice::smGFXDevice = NULL;
 bool GFXDevice::smWireframe = false;
+
 bool GFXDevice::smDisableVSync = true;
 bool GFXDevice::smPrevDisableVSync = GFXDevice::smDisableVSync;
+S32  GFXDevice::smFullscreenVSyncMode = VS_Auto;
+S32  GFXDevice::smWindowedVSyncMode = VS_Auto;
+
 F32 GFXDevice::smForcedPixVersion = -1.0f;
 bool GFXDevice::smDisableOcclusionQuery = false;
 bool gDisassembleAllShaders = false;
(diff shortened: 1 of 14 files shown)


Engine/source/gfx/gfxDevice.cpp
```diff
@@ -71,15 +78,23 @@ void GFXDevice::initConsole()
       "them to return only the visibile state.\n"
       "@ingroup GFX\n" );
 
-   Con::addVariable( "$pref::Video::disableVerticalSync", TypeBool, &smDisableVSync,
+   Con::addVariable( "$video::disableVerticalSync", TypeBool, &smDisableVSync,
       "Disables vertical sync on the active device.\n"
-      "@note The video mode must be reset for the change to take affect.\n"
+      "@note The video mode must be reset for the change to take effect.\n"
```
eightyeight: :ok_hand:

Engine/source/windowManager/platformWindowMgr.h
```diff
@@ -136,6 +136,10 @@ class PlatformWindowManager
    /// This method indicates to created windows to show as normal.
    virtual void setDisplayWindow(bool set){}
 
+   /// Checks if DWM is running (Windows only)
+   virtual bool isDesktopCompositionEnabled() { return false; }
+   virtual void updateDesktopCompositionState() {}
```
eightyeight: Is there any way we could avoid adding Windows-specific capability to a general class?

bpay: The alternative is to dynamic_cast the window manager before using it in gfxDevice, but that seems worse to me - I would prefer not to add more reliance on RTTI. My thinking was that it is Windows-only for now, but there might be similar mechanisms on other platforms that people might want to slot in.

In general, this is the old problem of using inheritance hierarchies to model static polymorphism.

eightyeight: I'm all for not introducing more dynamic casts. Carry on!

Lopuska:
This is looking really functional and written in elegant way! (:
Thanks!

dottools:
Uh no. Gladly to accept the improvements, but changing the setting's namespace is a big no no. Leave it under `$Pref::Video::` where it belongs as it is a user preference after all.

Lopuska:
@bpay Yes, is there a specific reason you removed the pref:: scope?
If no, this could break the compatibility when you upgrade the engine.
Thanks @dottools for report.

Lopuska:
Well, I can include this PRs in my D3D9 refactor branch and restore namespace pref.
Thank you!

Lopuska:
I think D3D9 support will be dropped soon, so I don't know if this PR is still useful.

Lopuska:
Closing this as D3D9 will be soon deprecated in favor of OpenGL and DX11 renderer.

PR #1217, opened 2015-02-22
Delete allProjects.txt

eightyeight:
It's not even used.


eightyeight:
Ah. Turns out it's used in `generateAllProjects.bat` to decide which projects to generate. Seems like they should both go, to me. Does anyone actually use that script?

eightyeight:
Actually, I propose that instead of deleting these files we move them to `Tools/Project Generator`.

PR #1221, opened 2015-02-23
Taml implementation

lukaspj:
Soeh.. Yeah I'm not sure if this is something you want in stock or anything, I just thought I should clean up the branch and merge it with head.

So this implements Taml as described in [here](http://www.garagegames.com/community/blogs/view/22777) .
Things that needs discussing before this goes in: do people even want Taml?

Also, this includes minimal changes to the TinyXML library that is against general policy.. I didn't care enough to find a workaround for that :P 

Well now it's out there. Feel free to reject the PR or something like that, I know it's a bit controversial because of the previously mentioned issues. But, I just wanted you guys to have the option :P


eightyeight:
Stock colours are re-included?

Anyway, thanks for this! Looks like a lot of hard work. We'll definitely have to have some discussion about this. I do think something like TAML is in T3D's future, but in what form and how difficult it will be to get there... unknown.

JeffProgrammer:
:(
Long live .gui!

Hope this won't be a 3.7 thing :P

eightyeight:
It's very flattering of you to think we could get this pushed through in time for 3.7 :joy:

lukaspj:
Huh odd with the StockColor, I'll look into it.

Areloch:
While this was an exellent initial PR, and the foundation of the new PR I just pushed #1443, my new PR gets TAML, Assets and Modules in together. I'll go ahead and close this one.

But again, that new PR couldn't have been done without this initial work, so a huge thanks Lukas :)

PR #1223, opened 2015-02-23
DeferredBumpFeat order of operations corrections

Azaezel:


lukaspj:
Shouldn't this be vec2 and not float2?
https://github.com/GarageGames/Torque3D/pull/1223/files#diff-f665c9749c6322ebfc893dfbb25be3b1R234

LuisAntonRebollo:
Thx :)

PR #1225, opened 2015-02-25
Vignette updated for OpenGL

J0linar:
+added Files for Full and Empty Templates
+updated vignette.cs file
+added vignetteP.glsl shader file


eightyeight:
Thanks! Good idea.

PR #1227, opened 2015-02-26
Con::executef trampoline had mismatched argc values for the high end

Azaezel:
Con::executef trampoline had mismatched argc values for the high end (caught by @jamesu)


eightyeight:
T_T

JeffProgrammer:
good catch az ;)

Azaezel:
mango caught it. just passing it along.

PR #1231, opened 2015-02-28
+proper buffer size for udp socket

J0linar:
SkypeLog: 28.02.2015
[17:18:32] Dušan Jocic: when creating user datagram protocol or UDP, old
implementation was limiting lan utalization to 1% of our modern 1000Mbps
hardwar
[17:19:15] Dušan Jocic: that add proper buffer size for udp socket
[17:19:43] Dušan Jocic: and lift off limitation what was present there
for 10/100 mbps lan

### DIFF
--- a/Engine/source/platform/platformNet.cpp
+++ b/Engine/source/platform/platformNet.cpp
@@ -503,7 +503,7 @@ bool Net::openPort(S32 port, bool doBind)
 	  }
 
       if(error == NoError)
-         error = setBufferSize(udpSocket, 32768);
+         error = setBufferSize(udpSocket, 32768*8);
 
       if(error == NoError && !useVDP)
          error = setBroadcast(udpSocket, true);


Azaezel:
just to be 100% sure: how would this one interact with https://github.com/GarageGames/Torque3D/pull/1001 ? 
(note gPacketRateToClient and gPacketSize have script correspondences to $Pref:: entries, and it's been common practice to bump those up this end at least for a while now)

dottools:
This PR just increases OS-side managed socket send and receive buffers. It won't have any bad effect on Torque's assembled packet size and xfer rates.

Lopuska:
As @dottools said, this just increase internal socket buffer 8 times. It's ok, but:
32k * 8 = 256k.
Why 256k?
And does this resolve the limitation about Torque3D usage in a gigabit lan network?

Lopuska:
Well, I thought carefully about that. It's ok.
Maybe it would be nice for the future to expose this param on prefs or in some kind of static/global var accessible via console script for easier future tweaks.
Thanks for contribution.

PR #1236, opened 2015-03-04
Revert recent style cleanup changes

eightyeight:
It seems to be fairly unanimous that the style cleanup PR was a bad idea, especially so late in 3.7, so we'll revert it and re-add it for 3.8, subject to a little more discussion. In the meantime, get thee gone!

Note that this means the 3,000 line changeset will appear three times in the history - original merge, reversion, and reinstatement (if we decide we want it in the end). Sorry, but that's just how the cookie crumbles. Rebasing history in a public repo is a can of worms I'm not gonna open right now.

### DIFF
--- a/Engine/source/T3D/debris.cpp
+++ b/Engine/source/T3D/debris.cpp
@@ -573,7 +573,7 @@ bool Debris::onAdd()
    // Setup our bounding box
    if( mDataBlock->shape )
    {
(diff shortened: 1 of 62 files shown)


JeffProgrammer:
stuff happens, honestly think its better to do that instead of hoping it don't break more stuff :)

PR #1239, opened 2015-03-05
Fix errors flagged by address sanitizer

bpay:


Engine/source/platformWin32/winFileio.cpp
```diff
@@ -1306,8 +1306,10 @@ static bool recurseDumpDirectories(const char *basePath, const char *subPath, Ve
    // Compose our search string - Format : ([path]/[subpath]/*)
    //-----------------------------------------------------------------------------
 
-   char trail = basePath[ dStrlen(basePath) - 1 ];
-   char subTrail = subPath ? subPath[ dStrlen(subPath) - 1 ] : '\0';
+   dsize_t trLen = basePath ? dStrlen(basePath) : 0;
+   dsize_t subtrLen = subPath ? dStrlen(subPath) : 0;
+   char trail = trLen > 0 ? basePath[ trLen - 1 ] : '\0';
+   char subTrail = subtrLen > 0 ? subPath[ subtrLen - 1 ] : '\0';
```
eightyeight: Props for using `dsize_t`.

Engine/source/gui/controls/guiTextCtrl.cpp
```diff
@@ -188,7 +188,7 @@ void GuiTextCtrl::setText(const char *txt)
    if( !mProfile )
       return;
    
-   if (txt)
+   if (txt && txt != mText)
```
eightyeight: Is that pointer comparison? Surely neither of these are `StringTableEntry`s so wouldn't a string comparison be more appropriate? And if so, it begs the question whether it's worth doing a comparison to avoid a copy :P.

dottools: Indeed, no point doing a string compare if all you're doing is avoiding a string copy since the string comparison will incur more overhead than the copying.

bpay: It's a pointer comparison, not a string compare. strncpy was being called with identical src and dest pointers, which is a no-no according to the C standard (although it happens to work with most implementations). I believe this happens when GuiTextEditCtrl calls setText(mText). I wasn't sure what the purpose of that was, but I assume it was being done for a reason, so I added the pointer check in order to avoid the violation and error report by asan.

eightyeight: Right, I thought of that afterwards then thought 'nah that's crazy, who'd ever do that!' Maybe add a comment for future generations?

bpay:
Added a comment about the overlapping pointers

PR #1241, opened 2015-03-07
vignette_final

J0linar:
+added: gui slider for maximum exposure of vignetting
+changed: shaders to logical min/max floats

This pr contains the glsl shaders as well and therefore makes the
https://github.com/GarageGames/Torque3D/pull/1225 Pull Request obsolete


eightyeight:
@LuisAntonRebollo I think this should go in 3.7 because the DX vignette shader is going in 3.7, so I'd feel better having both DX and OpenGL support in 3.7. Looks like it'll be easy to test.

LuisAntonRebollo:
oh, i forgot we added this for HLSL.

Thx :)

PR #1242, opened 2015-03-07
Fix Linux rpath.

LuisAntonRebollo:
Necesary for distribute binaries.

### DIFF
--- a/Tools/CMake/torque3d.cmake
+++ b/Tools/CMake/torque3d.cmake
@@ -29,8 +29,8 @@ if(UNIX)
     #set(CXX_FLAG32 "-m32") #uncomment for build x32 on OSx64
     
     # default compiler flags
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXX_FLAG32} -Wall -Wundef -msse -pipe -Wfatal-errors ${TORQUE_ADDITIONAL_LINKER_FLAGS} -Wl,-rpath=.")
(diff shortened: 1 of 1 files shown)


eightyeight:
@bpay (I think it was) advised that `rpath` should be set to `$ORIGIN`, not `.` for security reasons. I'm not fully aware of the differences, though.

LuisAntonRebollo:
Changed.

I have curiosity about diference. Anyone can post some link with info? Plz :P

bpay:
see discussion on #1158

LuisAntonRebollo:
wow... i miss some important talk :P

Thx a lot.

eightyeight:
Thanks Ben. Yeah it failed when I tried it as well. For now, maybe we should rely on documentation, or run scripts that provide the correct path at runtime. Can anyone verify that IDEs aren't doing something about this when you build with them? I've only built from the command-line. AFAIR Terminal Overload is providing users with a script they run instead of the executable directly. @fr1tz could comment.

fr1tz:
I assume using "." instead of "$origin" causes the linker to search for the libs relative to the current working dir. If that's the case it would be a potential security risk (a bit like having "." in your PATH).
For Terminal Overload releases I use '$ORIGIN/lib' as rpath (set using the "chrpath" utility). The script @eightyeight mentioned just checks if a 32-bit version of libGL is available, because that's the only dependency not included in the release distribution (which even includes a 32-bit version of Linux' dynamic linker). The script is located at https://github.com/fr1tz/terminal-overload/blob/master/My%20Projects/main/game/overload.sh and it seems to work. Or at least there's a lack of feedback that it doesn't work ;)

LuisAntonRebollo:
Fixed Clang.

Issue #1135, opened 2015-01-28
Object editor lod alteration fails


Azaezel:
http://i.imgur.com/c6npZqc.jpg

GuiTreeViewCtrl::getItemValue: invalid item id!
TSShapeConstructor::getMeshType: Could not find mesh 'detail3'
TSShapeConstructor::getObjectNode: Could not find object 'detail'
TSShape::setMeshSize: Could not find mesh 'detail 3'

This is both with, and without attempting to resolve via the suggested reversion https://github.com/GarageGames/Torque3D/issues/1134

jamesu:
Does it work before the revision?

Azaezel:
nope. same error either way.

jamesu:
Ok just checking if I broke anything ;)

LuisAntonRebollo:
@Azaezel , can you test if this PR #1246 is correct?

Fixes error msg and looks a correct fix, but not sure.

PR #1246, opened 2015-03-08
Fix GuiTreeViewCtrl::getParentItem incorrent use in ShapeEditor script files.

LuisAntonRebollo:
Related to  Object editor lod alteration fails #1135 

If you check script files, use ```%this.getParent( %id );``` but are a invalid method:
```
DefineEngineMethod( GuiControl, getParent, GuiControl*, (),,
   "Get the immediate parent control of the control.\n"
   "@return The immediate parent GuiControl or 0 if the control is not parented to a GuiControl.\n" )
{
   return object->getParent();
}
```

Looks like the correct method is:
```
//FIXME: [rene 11/09/09 - This clashes with GuiControl.getParent(); bad thing; should be getParentItem]
DefineEngineMethod( GuiTreeViewCtrl, getParentItem, S32, (S32 itemId), ,
   "Get the parent of a given id in the tree.\n\n"
   "@param itemId TreeItemID of item that has a parent we should get.\n"
   "@return Id of parent of given item.")
{      
   return(object->getParentItem(itemId));
}
```

### DIFF
--- a/Templates/Empty/game/tools/shapeEditor/scripts/shapeEditor.ed.cs
+++ b/Templates/Empty/game/tools/shapeEditor/scripts/shapeEditor.ed.cs
@@ -1132,7 +1132,7 @@ function ShapeEdNodes::onDeleteNode( %this )
 function ShapeEdNodeTreeView::getChildIndexByName( %this, %name )
 {
    %id = %this.findItemByName( %name );
-   %parentId = %this.getParent( %id );
+   %parentId = %this.getParentItem( %id );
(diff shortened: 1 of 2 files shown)


eightyeight:
Awesome work, thanks @LuisAntonRebollo! Will test tonight.

Azaezel:
verified. both lod and crash errors fixed.

eightyeight:
:grinning:

eightyeight:
Or, you know, I'll check it now.

PR #1259, opened 2015-03-18
early out when attempting to get a NULL index entry.

Azaezel:
early out when attempting to get a NULL index entry.

### DIFF
--- a/Engine/source/sim/netGhost.cpp
+++ b/Engine/source/sim/netGhost.cpp
@@ -1195,6 +1195,9 @@ NetObject *NetConnection::resolveObjectFromGhostIndex(S32 id)
 
 S32 NetConnection::getGhostIndex(NetObject *obj)
 {
+   if (!obj)
+      return -1;
+
    if(!isGhostingFrom())
       return obj->mNetIndex;
    S32 index = obj->getId() & (GhostLookupTableSize - 1);


Engine/source/sim/netGhost.cpp
```diff
@@ -1195,6 +1195,9 @@ NetObject *NetConnection::resolveObjectFromGhostIndex(S32 id)
 
 S32 NetConnection::getGhostIndex(NetObject *obj)
 {
+   if (!obj)
```
jamesu: Every call to this method I can see seems to check if obj is null. Maybe it would be a better idea to have an AssertFatal in here. Either that or change every instance of this call so a redundant check isn't performed.

dottools:
The real bug is whatever is calling `getGhostIndex()` with a NULL pointer. Perhaps it should be a fatal assert instead to spot and fix those bugs instead as James Urquhart suggested on IRC.

Azaezel:
Simple reproduction case: one object holding onto or throwing out a pointer to another:
::packUpdate(NetConnection *con, U32 mask, BitStream *stream)
{...
   if (mLinkage != mPrevLinkage) {
      S32 gIndex = con->getGhostIndex(dynamic_cast<ShapeBase*>(mLinkage));
      if (stream->writeFlag(gIndex != -1)) {
         stream->writeInt(gIndex, NetConnection::GhostIdBitSize);
      }
      mPrevLinkage = mLinkage;
   }
   else
      stream->writeFlag(false);
 
   return retMask;
}
 
::unpackUpdate(NetConnection *con, BitStream *stream)
{...
   if (stream->readFlag()) {
      S32 gIndex = stream->readInt(NetConnection::GhostIdBitSize);
      mLinkage = dynamic_cast<ShapeBase*>(con->resolveGhost(gIndex));
   }
}

could see further altering the pr to do a dynamic cast in-method. doesn't however make a great deal of sense to require functionality for a method to function that is not part of the method itsself.

eightyeight:
Should the method in fact take a handle, not a pointer? :P

PR #1253, opened 2015-03-14
Two equal lines in init.cs

John3:
this: http://forums.torque3d.org/viewtopic.php?f=11&t=113

PR #1264, opened 2015-03-26
Two equal lines in init.cs

John3:
this is only present in the full template.

Resubmission of #1253 - @eightyeight


eightyeight:
Usually, we create a new branch for each PR, which means you won't accidentally add commits to it. We also recommend that if there's work you don't plan on contributing back to the main repo, that you maintain your own long-term branch for it (not `development`), so that you can just pull down updates directly into `development` and merge from there.

PR #1282, opened 2015-04-23
buttoninspector field

Azaezel:
extract from @andr3wmac:
adds the capacity to include an execution button via initpersistfields to tie that automatically into editors

primary example: http://i.imgur.com/oXr22sU.jpg

### DIFF
--- a/Engine/source/console/consoleObject.h
+++ b/Engine/source/console/consoleObject.h
@@ -460,6 +460,7 @@ public:
    enum FieldFlags
    {
       FIELD_HideInInspectors     = BIT( 0 ),    ///< Do not show the field in inspectors.
+      FIELD_ButtonInInspectors   = BIT( 1 ),    ///< Display as a button in inspectors.
    };
 
    struct Field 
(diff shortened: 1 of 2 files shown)


Azaezel:
https://github.com/Azaezel/Torque3D/blob/PBR_LPV/Engine/source/T3D/offlineLPV/offlineLPV.cpp#L178-L179 (Tie in usage sample. know at least one other project in the works that could likely make use of that, as well as a couple others that spring to mind, like walkabout.)

eightyeight:
I used a similar trick in the recast resource (pre Walkabout) with a checkbox AFAIR, only because I was too lazy to build an editor mode. IMO this is a hack, but it's a small and localised one, so why not. Any chance we could add a new `addButton` method to replace `addProtectedField`? That way we could maybe refactor the behaviour at some point to, I dunno, put all the buttons in their own dedicated panel instead of inline with fields?

Azaezel:
@eightyeight: that was actually the topic of IRC two days ago. I believe @Areloch mentioned having an alternate that would be suitable for say, slipping in to support the revised interface needs for https://github.com/GarageGames/Torque3D/pull/750 in a more flexible manner and the like. *coughpokereminderpokecough*

Azaezel:
variation added via e/c experimental

Issue #1265, opened 2015-03-31
No OpenAL output in Debian Wheezy


Levitator1:
Reposted as requested:

It builds! It runs! However, there is no sound on Debian Wheezy. During startup, I see this in the console:

SFXSystem::createDevice - created OpenAL device 'OpenAL Soft'
Provider: OpenAL
Device: OpenAL Soft
Hardware: No
Buffers: 16

I hear a brief crackle as this scrolls by, but when playing the demo maps, there is no sound at all. It was just now suggested to me to try FMOD instead.

JeffProgrammer:
afaik the FMOD implementation in torque is the old one, and requires the use of the legacy version of fmod, not the new one with the indie license :(

eightyeight:
Just realised this was fixed.

PR #1290, opened 2015-04-28
Patch to fix no-sound in Linux

Levitator1:
The default multi-threaded build under Linux had no sound, and this patch fixes that. There was a problem with the way that worker threads are awakened, and the background loading of sounds was not happening.

Addresses [Issue #1265](https://github.com/GarageGames/Torque3D/issues/1265)


eightyeight:
Thanks for this contribution! Sorry we haven't gotten back to you before now. I'll be testing this soon. Just a note for the future - as we're using git, we prefer not to leave commented-out code in the repo. The revision history will show that information without cluttering the code. Also, we prefer comments that explain the current situation, rather than having a discussion with past developers. I'll edit that out when I merge your commit, just so you know!

eightyeight:
Merged in d4e5b471. Thanks!

PR #1294, opened 2015-05-01
opengl crashfix: cannot self-multiply a uniform. use a temp-variable.

Azaezel:
opengl crashfix: cannot self-multiply a uniform. use a temp-variable. (directx altered for parity)


eightyeight:
Merged in 462aeef. Thanks!

PR #1295, opened 2015-05-02
Add check for empty string before overwriting the object data file name

RichardRanft:
The EngineAPI method passes in "" by default - which is not null, so passes the if(path) condition and overwrites the object's mDataFileName with an empty string.  This in turn asserts when it tries to save the file.

So, adding a check for an empty string seems the "cheapest" fix.

### DIFF
--- a/Engine/source/forest/forest.cpp
+++ b/Engine/source/forest/forest.cpp
@@ -354,7 +354,7 @@ void Forest::createNewFile()
 
 void Forest::saveDataFile( const char *path )
 {
-   if ( path )
+   if ( path && strcmp(path, "") != 0)
       mDataFileName = StringTable->insert( path );
 
    if ( mData )


dottools:
You cannot use `strcmp()` to compare empty strings. Only `strlen() == 0` can determine that.

bpay:
> You cannot use strcmp() to compare empty strings. Only strlen() == 0 can determine that.

According to whom?

The idiomatic way to do this is `if (str && *str)`

eightyeight:
Did we not add a [utility function](https://github.com/GarageGames/Torque3D/blob/1c580f5a5f2fc93a7432316275ae394d71dfc519/Engine/source/core/strings/stringFunctions.h#L63) for this in a prior PR? Abstraction, people!

RichardRanft:
It works fine - I've tested it both passing an empty string and passing it a valid file name.  So if it can't be done that way then I'm the exception.  However, it does seem that the mission likes to keep it named a certain way so, while the new file name will be accepted and the forest data will be saved there the mission with then happily save the <mission name>.forest file name for the forest object in the .mis file.

eightyeight:
`strlen` and `strcmp` methods are both _O(n)_ for the record, not that it will matter here. I would prefer using the empty string utility method simply for ease of reading.

Areloch:
Handled in #1299

Issue #1296, opened 2015-05-06
Oculus VR DK2 Support


jamesu:
The Engine needs support for the latest DK2 and 5.x SDK. Posting an issue here so it can be better tracked.

Also an update to the wiki documentation would be useful as currently it is awful.

jamesu:
Submitted a PR

Areloch:
Merged the PR.

PR #1297, opened 2015-05-06
Oculus VR DK2 Support

jamesu:
- Updated to work with 0.5.x SDK
- Uses Oculus Rendering rather than PostFX
- Stereo rendering refactored so more rendering info is grabbed from the DisplayDevice
- Implements an Offscreen Canvas for in-game gui with oculus
- Message dialogs and metrics display can now go to the OffScreen Canvas (if oculus demo is setup correctly)

Resolves issue #1296


Areloch:
Having my friend double-test this as he isn't half-blind in one eye without glasses. Should have his confirmation that it works as well as I thought it did tomorrow. If so, I'll merge this in.

eightyeight:
:grinning:

Areloch:
'Aight, my friend tested it for me.

He said, like we saw, rendering seems to be fine, but there's that little bit of an interpolation jitter that needs to get sorted out going forward.

He also said that the starting prompt when a level loads looks fine, but if you hit escape to get the exit out ui pop-up, that's overshot. "Left eye: box too far right, Right eye: box too far left". So the respective eye's offsets for the UIs are overshot. So that'll need a tweak.

But otherwise, the root implementation seems to be solid. So if there are no immediate complaints, we can merge this in, and then patch those remaining holes in follow-up PRs.

Sound reasonable?

jamesu:
@Areloch were you running in debug or release? Release tends to be much smoother.

The jitter itself I think is caused by the different values returned from the sensor data which gets sent to the move manager and the data which is sent by the per-frame rendering callbacks.

You can force it to only use the data from the move manager by setting "$TSControl::useLatestDisplayTransform" which will give more consistent movement, but this may cause problems with the time warp distortion especially if the FPS is too low.

From what I can tell most examples tend to either use the rendering transform as-is OR they interpolate it with the transform of the control object. The additional problem with torque is in order to properly update the view frustum the movement needs to go through the MoveManager. On top of that you have additional movement caused by using the keyboard or mouse to move the player. Also to be honest the Oculus SDK isn't entirely clear on how these problems should be resolved in a real situation (though I did hear they changed the API yet again in 0.6 so I'll have to check if thats improved).

Rendering-wise this adds some crucial features for properly rendering stereo viewports with asymmetric  projections which would be useful for other VR devices.

Areloch:
Thinking on it, I may have passed him the debug build. So if it's even less of a problem in release, I won't fret on it majorly for now.

And I agree that this introduces stuff not only for oculus, but other VR systems, so it's definitely good to have. 

So we can merge this in now, and just continue to polish the remaining oddities going forward.

eightyeight:
:grinning:

Issue #1289, opened 2015-04-27
Dedicated server crashes


eightyeight:
3.7 release candidate is showing some crashes with the dedicated server as reported [here](http://www.garagegames.com/community/blogs/view/22963/1#comment-199966). Symptoms in Linux are slightly different - the dedicated server console is unresponsive, but the engine still functions normally, and can act as a host.

eightyeight:
Problem is in here:

```cpp
void postConsoleInput( RawData data )
{
   // Schedule this to happen at the next time event.
   ConsoleValueRef argv[2];
   argv[0] = "eval";
   argv[1] = ( const char* ) data.data;
   Sim::postCurrentEvent(Sim::getRootGroup(), new SimConsoleEvent(2, argv, false));
}
```

The first assignment causes an assert in `ConsoleValueRef::operator=`. @jamesu? This pattern seems to be used in a couple of places. Wondering if this is a correct use of `ConsoleValueRef`. The comment on the class indicates:

```cpp
// Can point to existing console variables,
// or act like a free floating value.
```

But I see no ways to construct a new `value` for it, only to assign it from an existing `ConsoleValue`.

eightyeight:
Further, for example, this constructor:

```cpp
ConsoleValueRef::ConsoleValueRef(const char *newValue) : value(NULL)
{
   *this = newValue;
}
```

Will cause this to happen immediately:

```cpp
ConsoleValueRef& ConsoleValueRef::operator=(const char *newValue)
{
   AssertFatal(value, "value should not be NULL");
   value->setStringValue(newValue);
   return *this;
}
```

...won't it?

jamesu:
Indeed, the ConsoleValueRef should point to a ConsoleValue. e.g.

	ConsoleValue values[2];
	ConsoleValueRef refValues[2];

	values[0].init();
	values[0].setStringValue("eval");
	values[1].init();
	values[1].setStringValue((const char*)data.data);
	refValues[0].value = &values[0];
	refValues[1].value = &values[1];

This behaviour was previously changed due to the previous issues regarding storing new stack values in ConsoleValueRef.

eightyeight:
Okay, I get it. I figured something like that was the case, but the intent of the code is unclear. How do you feel about my second comment? Is the `const char*` constructor a bug?

jamesu:
Encountering this issue with the oculus stuff in a windowed console. Is anyone making a patch or should I do it?

eightyeight:
[...]

PR #1298, opened 2015-05-06
Remove value constructors for ConsoleValueRef & fix callbacks

jamesu:
Fixes issue #1289

### DIFF
--- a/Engine/source/console/simEvents.cpp
+++ b/Engine/source/console/simEvents.cpp
@@ -129,7 +129,7 @@ ConsoleValueRef SimConsoleThreadExecCallback::waitForResult()
       return retVal;
    }
 
-   return (const char*)NULL;
+   return ConsoleValueRef();
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Engine/source/console/simEvents.cpp
```diff
@@ -129,7 +129,7 @@ ConsoleValueRef SimConsoleThreadExecCallback::waitForResult()
       return retVal;
    }
 
-   return (const char*)NULL;
+   return ConsoleValueRef();
```
eightyeight: As far as I can tell, this should never be called, right?

eightyeight:
Will merge this manually after review.

Engine/source/console/simEvents.cpp
jamesu: I'm guessing no but it doesn't make sense to return retVal if for whatever reason acquire fails.

eightyeight: Yep, I was just trying to understand the flow.

eightyeight:
Yay it works. Thanks @jamesu! Merged in 12d435c.

PR #1299, opened 2015-05-06
Add static String::isEmpty(const char*) method

RichardRanft:
Add static String::isEmpty(const char*) method and use it to verify path for Forest::saveDataFile() is not empty.  Personally not a fan of the old, free-floating dStrBlahBlah functions - I think they would be easier accessed (and remembered) as String::BlahBlah() static methods, especially since intellisense will help you out this way.

This is preferable to creating a String from the const char* and then asking the new string if it's empty.

Consider it an "improvement" on https://github.com/GarageGames/Torque3D/pull/1295

### DIFF
--- a/Engine/source/core/util/str.cpp
+++ b/Engine/source/core/util/str.cpp
@@ -659,7 +659,7 @@ bool String::isEmpty() const
 
 bool String::isEmpty(const char* str)
 {
-	return (dStrlen(str) == 0);
+	return src == 0 || src[0] == '\0';
 }
 
 bool String::isShared() const
--- a/Engine/source/core/util/str.cpp
+++ b/Engine/source/core/util/str.cpp
@@ -657,6 +657,11 @@ bool String::isEmpty() const
    return ( _string == StringData::Empty() );
 }
 
+bool String::isEmpty(const char* str)
+{
+	return (dStrlen(str) == 0);
+}
+
 bool String::isShared() const
 {
    return _string->isShared();
--- a/Engine/source/core/util/str.cpp
+++ b/Engine/source/core/util/str.cpp
@@ -659,7 +659,7 @@ bool String::isEmpty() const
 
 bool String::isEmpty(const char* str)
 {
-	return src == 0 || src[0] == '\0';
+	return str == 0 || str[0] == '\0';
 }
 
 bool String::isShared() const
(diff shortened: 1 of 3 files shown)


Areloch:
Only thing I'd change here is to have the code in the function match our current dStrIsEmpty

`return src == 0 || src[0] == '\0';`

Otherwise, this looks good. If you could make that tweak on this Richard I think we could probably roll this in.

eightyeight:
:+1:

Areloch:
Alright, got a chance to test this.

2 changes to make:

first, instead of src == 0 || src[0] == '\0', it should be str, as per the passed in parameter. 

Second, if we fail that check and we thus don't have a mDataFile set(it being a blank string), we may still have mData set from before, and it'll try to save to a blank file, crashing.

Would it make more sense to make it return out if the path is empty? Or keep the asset so you aren't trying to save to blank files?

RichardRanft:
Re: second point - 
The reason this was crashing before is that the previous behavior simply tried to save to whatever filename was passed and it is "" by default.  I believe the original simply checked to see if path was null and replace the current mDataFileName if it is - but since it's "" it is NOT null, so it tries to replace mDataFileName with "" and then attempts to save to that name.  I mDataFileName is set when the object is added to the scene (see Forest::onAdd() - it calls createNewFile(), which gives the forest a default filename) and should not be empty, so the new behavior will use the value stored in mDataFileName at creation if path is null OR "", otherwise try to use the (not null, not empty string) path provided.

So, if the forest object exists it has a valid file name - we just want to prevent replacing the valid name with an invalid one.

Areloch:
Well, I understood the issue, I just wasn't sure if it made more sense to just not enact the save if the path was blank, or leave the assert in to dissuade bad behavior.

But, requiring them to use a valid file or not at all makes total sense, so hey whatevs. With the src->str correction, this looks good. So i'll merge it in.

RichardRanft:
But my point is, calling it with an empty path is simply saying "use the path in mDataFileName."  It's a convenience - since there will always be a file path present to use and therefore it won't crash with an empty path in obj->write(), so it's now harmless.  I'm guessing this broke when someone "fixed" the console method to use "" as default (which seemed reasonable until one realizes that now path isn't null and "" is a valid string, if not a valid path).

PR #1301, opened 2015-05-09
setExtent now takes Strings instead of leaving args as pointers.

### LINKED ISSUE #1292: Editor ribbon cannot be expanded once collapsed
Clicking the little arrow on the ribbon to collapse the icons works, but clicking it again to expand them fails with a console message. EDIT: introduced in #1092, I think. EDIT: reverting 61e106251 made no difference. Hmm.

eightyeight:
It appears that passing raw expressions as arguments (e.g. `setExtent(..exp.., 4)`) was causing the `extOrX` value to be the same as `y`, which was unhelpful and bizarre. This fixes it, but I'm not sure how. Fixes #1292.


eightyeight:
Closed in favour of #1303.

PR #1303, opened 2015-05-15
Add workaround for issue #1292

jamesu:
Works around the problem of string conversions for integer and float arg values by temporarily creating a string.

### DIFF
--- a/Engine/source/console/console.cpp
+++ b/Engine/source/console/console.cpp
@@ -1769,7 +1769,7 @@ const char *ConsoleValue::getStringValue()
       else if(type == TypeInternalInt)
          internalValue = Con::getData(TypeS32, &ival, 0);
       else
-         internalValue = Con::getData(type, dataPtr, 0, enumTable);
+         return Con::getData(type, dataPtr, 0, enumTable); // We can't save sval here since it is the same as dataPtr
 
       if (!internalValue)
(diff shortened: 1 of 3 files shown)


eightyeight:
When I run the engine, I get an immediate crash in VS2012 Windows 7 x64 on line 1760:

```cpp
         sval = (char *) dRealloc(sval, newLen);
```

Same happens on Ubuntu 14 with gcc. It appears to happen while exec'ing a `material.cs`.

jamesu:
Hmm odd, taking a look

jamesu:
Not seeing a problem here. Could you be more specific with your reproduction instructions?

jamesu:
Checked wrong branch, whoops

jamesu:
Fixed. Registered console values use the dataPtr variable which conflicts with sval.

Also tidied up the handling of sval so it only checks bufferLen instead of deducing its allocated based on type.

eightyeight:
Checks out on Ubuntu, checking Windows.

Areloch:
Looks like it works for me as well.

eightyeight:
TODO: merge into `release-3.7`. Need to cherry-pick instead of merging directly.

eightyeight:
Cherry-picked in d9436e6

Issue #1304, opened 2015-05-16
Reset PostFX settings between every mission


eightyeight:
Currently, if you enter a mission with some custom PostFX preset (`P1`), which includes settings for some PostFX `FX1`, then you exit that mission and load another mission with its own custom presets (`P2`), which don't include any settings for `FX`, then you get all of `P1`'s settings in this mission, instead of the defaults.

I propose that the templates should exec `core/scripts/client/postFx/default.postfxpresets.cs` before _every_ mission load, _then_ a custom presets file if one is available, to ensure consistent behaviour when new effects are added to the engine, but not to everyone's mission preset files.

Might also be worth thinking about having a separate defaults file, or allowing you to save the current settings to the defaults for your game.

PR #1305, opened 2015-05-16
Apply vignette settings properly for #1302

### LINKED ISSUE #1302: Vignette PostFX does not save
The Vignette PostFX does not save it's values to the postfx.preset file or anywhere else, only on or off gets saved, the strength slider will get reset on game reload.
At the same time the setting for vignette will get carried over to other levels when you open them, even if you did not enable vignette for them.

https://github.com/GarageGames/Torque3D/commit/0e346fed9ffd509b205ab9cba73e0fb34ca902ad

As far as I can see it uses a variable, but it does not get saved anywhere as a preference or so and it does not have variables that get saved in the postfx presets except the on/off button.
No idea how it is supposed to work, but to me it looks like it should be possible to define the vignette values into postfx presets for each mission.

eightyeight:
@Duion see if this fixes #1302 for you. Seems to work fine for me!

Annoyingly, there's no cleanup before running a level-specific `postfxpreset.cs` file, so each of these files have to have settings for _all_ postFX in them, hence the change in the Outpost presets (to turn the vignette off). See #1304.

### DIFF
--- a/Templates/Empty/game/core/scripts/client/postFx/default.postfxpreset.cs
+++ b/Templates/Empty/game/core/scripts/client/postFx/default.postfxpreset.cs
@@ -20,11 +20,13 @@
 // IN THE SOFTWARE.
 //-----------------------------------------------------------------------------
 
+$PostFXManager::Settings::EnableVignette = "0";
 $PostFXManager::Settings::EnableDOF = "0";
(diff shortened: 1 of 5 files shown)


Areloch:
Looks good to me. If there are no other complaints, we can merge this in.

Issue #1287, opened 2015-04-27
Forest in Outpost level not appearing in Ubuntu


eightyeight:
Level loads, but no forest trees exist. Relevant extract from `console.log`:

```
Executing levels/Outpost.mis.
Failed to create resource: [levels/outpost.forest]
levels/Outpost.mis (0): Register object failed for object theForest of class Forest.
```

Putting this here to remind myself to look into it, and see if anyone else has the issue.

LuisAntonRebollo:
Can be a case-sensitive problem in the file name.

```
Executing levels/Outpost.mis.
Failed to create resource: [levels/outpost.forest]
```

eightyeight:
:sob: I'll check it later on.

eightyeight:
Okay, that was a problem, but a separate one. Edited the post to remove those parts - only the first three lines are relevant to the forest.

eightyeight:
@Azaezel, you were right. When you posted that line with the `defaulttree.dae`, I thought you were referring to the _other_ place that line exists, which is in some random TSStatic that was also causing console errors, which I had fixed, but hadn't fixed in the forest item definition.

As for the source of the crash, the problem originates in [tsForestItemData.h](https://github.com/GarageGames/Torque3D/blob/b30b5d907623a89ddb812992de1b0c588f0f41cf/Engine/source/forest/ts/tsForestItemData.h#L91):

```cpp
   const Box3F& getObjBox() const { return mShape ? mShape->bounds : Box3F::Invalid; }
```

When the shape file it not found, every item gets given an invalid object box. Then, when the forest cell goes to [update its bounds](https://github.com/GarageGames/Torque3D/blob/b30b5d907623a89ddb812992de1b0c588f0f41cf/Engine/source/forest/forestCell.cpp#L180), it tries to update with all these invalid boxes.

I've made two PRs from this. #1307 fixes the default bounds issue. If you're missing a shape file, you really don't want a crash, you want a console error, which you do currently get. #1308 fixes the filename case :p.

Oh, and #1309 adds more detail to that fatal assert we were looking at :p.

PR #1307, opened 2015-05-16
tsForestItemData: default to no bounds instead of crazy bounds

eightyeight:
See #1287.

> As for the source of the crash, the problem originates in [tsForestItemData.h](https://github.com/GarageGames/Torque3D/blob/b30b5d907623a89ddb812992de1b0c588f0f41cf/Engine/source/forest/ts/tsForestItemData.h#L91). When the shape file is not found, every item gets given an invalid object box. Then, when the forest cell goes to [update its bounds](https://github.com/GarageGames/Torque3D/blob/b30b5d907623a89ddb812992de1b0c588f0f41cf/Engine/source/forest/forestCell.cpp#L180), it tries to update with all these invalid boxes.
>
> If you're missing a shape file, you really don't want a crash, you want a console error, which you do currently get.

### DIFF
--- a/Engine/source/forest/ts/tsForestItemData.h
+++ b/Engine/source/forest/ts/tsForestItemData.h
@@ -88,7 +88,7 @@ public:
    const Vector<S32>& getLOSDetails() const { return mLOSDetails; }
 
    // ForestItemData
-   const Box3F& getObjBox() const { return mShape ? mShape->bounds : Box3F::Invalid; }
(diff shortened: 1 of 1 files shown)


eightyeight:
Not merged immediately because I'd like a second opinion. Does zero-size box sound like a decent default if a shapefile isn't found? As noted, there is an error. I figure `Box3F::Invalid` must have been chosen for a reason.

Issue #1315, opened 2015-05-19
particle glow bug


Azaezel:
Directx shown: https://youtu.be/PZrZyB1PI5g uses https://github.com/GarageGames/Torque3D/pull/863 for debug viewing. Thoughts @lukaspj?

lukaspj:
That is sooo odd...

lukaspj:
If you've got the time, could you check that they are added here:
https://github.com/GarageGames/Torque3D/blob/30a8401d1461564a283810daacf78b5f6043a8d5/Engine/source/renderInstance/renderGlowMgr.cpp#L132-L136

And rendered here: https://github.com/GarageGames/Torque3D/blob/30a8401d1461564a283810daacf78b5f6043a8d5/Engine/source/renderInstance/renderGlowMgr.cpp#L184-L208
?

Azaezel:
Both trigger in debug mode, yes.

PR #1316, opened 2015-05-21
Fix for rendering particles to the glow buffer

lukaspj:
Fix for #1315

### DIFF
--- a/Engine/source/renderInstance/renderGlowMgr.cpp
+++ b/Engine/source/renderInstance/renderGlowMgr.cpp
@@ -202,6 +202,10 @@ void RenderGlowMgr::render( SceneRenderState *state )
          }
 
          ParticleRenderInst *ri = static_cast<ParticleRenderInst*>(_ri);
+
+         GFX->setStateBlock(mParticleRenderMgr->_getHighResStateBlock(ri));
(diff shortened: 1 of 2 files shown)


lukaspj:
Btw fun fact, the only occasion it would work earlier, was when the rendering order was such that it would render to the glow buffer exactly after rendering the particles :P Pure coincidence we didn't spot it right away :P

Azaezel:
works

eightyeight:
I'm wary of exposing methods that are evidently intended to be private. Can `RenderGlowMgr` be a friend of `RenderParticleMgr` so that we can at least contain the leak? Or is there a proper way to communicate the information?

lukaspj:
Meh, I'd say it doesn't make sense for those getter methods to be private anyways. But yeah we can make it a friend class.

eightyeight:
If they should be part of the public interface, can we make a ticket to make them properly public in the future? I.e. drop the underscore, and maybe move them to an already-public section?

Areloch:
Yeah, I think we can just do a cleanup pass on it in 3.8 or 3.9.

As for this, it looks good. If you didn't spot any problems Dan, I think we can roll this one in.

eightyeight:
Sure thing. Do the honours if you like, but you'll need to cherrypick it onto `release-3.7` since the PR is targeted at development.

eightyeight:
Cherry-picked in c5dd723.

eightyeight:
Cherry-picked in c5dd723.

PR #1321, opened 2015-06-06
No need for server-only aps to generate visible textures. 

Azaezel:
(#1320 had a transcription error)


eightyeight:
Merged in 6218d79. Thanks!

PR #1323, opened 2015-06-07
Replace uses of dStrCmp  with new String::compare

Areloch:
Moves from using dStrCmp to the new String::compare static functions keeps things cleaner, consistent, and works with intellisense.

This is the part of the move to replace usage of the old dStr___ functions for the more standard String namespace approach.

### DIFF
--- a/Engine/source/T3D/aiClient.cpp
+++ b/Engine/source/T3D/aiClient.cpp
@@ -403,7 +403,7 @@ void AIClient::onAdd( const char *nameSpace ) {
 
    // This doesn't work...
    //
-   if( dStrcmp( nameSpace, mNameSpace->mName ) ) {
+   if( String::compare( nameSpace, mNameSpace->mName ) ) {
       Con::linkNamespaces( mNameSpace->mName, nameSpace );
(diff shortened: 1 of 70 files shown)


Areloch:
As a note, I didn't notice any parts that actually used the int return in any way other than to determine if they were the same or different.

Given that, would it make sense to just have String::compare return a boolean? Or do we want to keep the int return just on the off chance someone may actually want to know where the difference is when using it?

eightyeight:
I reckon keeping the int is better. Maybe we should also have a `String::equal` function or something which returns a bool. But, you know, who cares about strong typing when we can cast from int to bool anywhere we want. :confused:

Areloch:
Yeah, I'd initially just figured 'we can just have an equals function' too, but when I was doing the convert work for this PR, I realized like 99% of the usages of compare were just testing if the return was 0 or not.

eightyeight:
Then they should use an equals function ;). Much less confusing that way, since `compare` logic is kind of backwards as compared to normal boolean logic.

Areloch:
Agreed. 

I'll do a separate PR after this one gets merged in to start converting those to a String::equals function instead.

Yay more readability!

PR #1325, opened 2015-06-08
standardization, and fix for editors with non-occulous builds (missing popups)

Azaezel:
\game\tools\worldEditor\scripts\editor.ed.cs function Editor::create() sets $GameCanvas = Canvas; with the rest following conversion conventions for https://github.com/GarageGames/Torque3D/pull/1297 . (see https://github.com/GarageGames/Torque3D/pull/1297/files#diff-e50af352c48fb4969d220ecd5a7521b9L79 on down)

 :

### DIFF
--- a/Templates/Full/game/tools/base/utils/swatchButtons.ed.cs
+++ b/Templates/Full/game/tools/base/utils/swatchButtons.ed.cs
@@ -37,7 +37,7 @@ function GuiSwatchButtonCtrl::onMouseDragged( %this )
    
    %xOffset = getWord( %payload.extent, 0 ) / 2;
    %yOffset = getWord( %payload.extent, 1 ) / 2;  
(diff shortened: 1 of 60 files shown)


eightyeight:
How can I reproduce the problem? I tested a couple of editor popups and they appear to work.

Azaezel:
At the time, the repro-check was adding and removing a terrain, and not receiving a popup. Must have been a transcription error this end somehow.

eightyeight:
I'll double check. So you think these changes are unnecessary? Or going to resubmit?

Azaezel:
Willing to say redundant unless similar lack of popups are otherwise reported, but will keep the fork around for say, a week or so just in case.

eightyeight:
Okie dokie, thanks for this.

PR #1327, opened 2015-06-14
GFXGLDevice setShader fix.

rextimmy:
Ok i stuffed up the other PR, branch was a mess. Anyways...attempt number 2 :)

This took ages to track down, compiling on VS 2013 was displaying nothing but the clear screen color when loading a level with GL. Long story short, the materials where not able to set their shader because the GFXGLDevice setShader function was wrong (missing the bool force).

### DIFF
--- a/Engine/source/gfx/gl/gfxGLDevice.cpp
+++ b/Engine/source/gfx/gl/gfxGLDevice.cpp
@@ -811,9 +811,9 @@ GFXShader* GFXGLDevice::createShader()
    return shader;
 }
 
-void GFXGLDevice::setShader( GFXShader *shader )
+void GFXGLDevice::setShader(GFXShader *shader, bool force)
 {
-   if(mCurrentShader == shader)
+   if(mCurrentShader == shader && !force)
       return;
 
    if ( shader )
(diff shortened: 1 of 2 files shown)


eightyeight:
Replaces #1326. Thanks for this! Glad you managed to solve the issue.

rextimmy:
One other nasty i noticed, when using #include with GLSL shader file, if you happen to have an #include "blah.h" within a comment it will try and load that too.

Azaezel:
Can confirm it works for non-occulous at least. Much obliged.

eightyeight:
Merged in b5f8e9a. Thanks!

Azaezel:
this needs rolling into development. that is where it's causing issues.

rextimmy:
Yeah sorry i only tested on development not release-3.7

Azaezel:
not exactly breaking things to have a never-triggered option added. but yeah. it's the former that's not displaying a thing and spewing errors in opengl.

eightyeight:
Ah right, sorry. I think we'll go ahead and release 3.7 tonight. I've been having trouble with the dev branch after merging in 3.7 and wanted to get that fixed before we released, but it seems like it's just me, so I shouldn't worry :p.

rextimmy:
I also never tested with GCC or VS < 2013 before the fix. I assume they where ok otherwise someone would have noticed long before.

eightyeight:
I think I tested it. Sounds like the sort of thing I'd do.

PR #1331, opened 2015-06-22
64bit Visual Studio Projects

Areloch:
This converts the Visual Studio projects generated by the Project Manager to default as 64 bit.


Areloch:
Note, don't know if 3.8 is fitting for this, or 3.9, but we'll be moving the default projects over to 64 bit at some point in the near-ish future, so having the PR exist now.

This only covers VS, the other templates need to be updated as well.

dottools:
Why not both 32bit and 64bit builds support?

eightyeight:
'Default', not 'only'. Ideal would be to work out the setup that Winterleaf had with 32 and 64 bit in a single project - though I think it's sadly not doable with CMake currently :/.

dottools:
@eightyeight: Can't tell if you're replying to Areloch or to me. But the changes clearly show that it's replacing all Win32 directives with x64, not appending 64bit configurations, but replacing 32bit with 64bit. So nothing defaulting about it since then you then have to manually create Win32 configurations in Visual Studio with these changes if you wanted Win32.

Areloch:
I mean, sure we can easily do both.

The question is how much merit there is to worrying about 32bit at this point. You have to practically go out of your way to not have a 64bit compliant machine - heck, even iOS has required 64bit support in EVERYTHING since Feb. - so I'm unsure if it's even something we'd target and just leave that up to the end user to retarget if they ABSOLUTELY require 32bit.

Thoughts?

eightyeight:
@dottools oh, when I'd discussed the issue with @Areloch I was under the impression that we were swapping the default, not replacing 32 with 64. Sorry about that. Jeff, I reckon we should definitely do both. I agree that 32 is a dwindling concern, but it's not 0 concern yet.

Areloch:
Sure, no problem.

If we wanna retain 32bit for a little while as a transitional period, that's no biggie. I'll update the PR with the 32bit projects back in there.

rextimmy:
Would be great if the project manager had a few new functions like addProjectLibInput64,addIncludePath64 etc so you can specify different libs/directories for the modules.

Areloch:
Given the shifting to using cmake as the backend, this is sorta unneded at this point, so I'm going to go ahead and close this PR.

PR #1332, opened 2015-06-24
initial aspect ratio maintaining relative gui scaling

Azaezel:
adds 3 new horizontal relative gui entries for dealing with swapping elements between 4:3 and 16:9 aspect ratios.  "relativeToYL" maintains the initial aspect ratio, keeping the leftmost edge the same, "relativeToYR" also shifts it so the right hand side is shifted over so that edge would match, and "relativeToYC" takes/adds space from both sides


Azaezel:
bide on this. requests for renames and augs.

Lopuska:
Tested and reviewed. Only corrected a missed letter in a comment.

Issue #1150, opened 2015-02-02
Fix all initializer list ordering


crabmusket:
There are a ton of warnings generated by `gcc`'s `-Wreorder`. The warning's meaning is explained [here](http://blog.kutny.net/2014/07/14/c-member-initializer-list-fallacy-a-k-a-wreorder/). Might not be any actual bugs, but it'd be nice to be on the safe side.

raistlin:
I have seen that there are a lot of -Wreorder warnings generated inside /lib/ files.
I think that is not a good idea to modify this files directly, because they will be overwritten if we update these libraries in future.

What do you think?

dottools:
Yeah, third party libraries shouldn't be modified unless absolutely necessary to integrate with the engine that otherwise wouldn't be useable.

eightyeight:
What @dottools said. I've mentioned before the possibility of using git submodules for our lib dependencies, at least those of them for which that's feasible - in that case maybe we could contribute back to those dependencies, maybe fix some warnings, etc. But that might not be a realistic ideal :p.

PR #1336, opened 2015-06-28
Fixed some minor compiler warnings on Linux

raistlin:
I am building this project on linux, gcc version 4.9.2 (Ubuntu 4.9.2-10ubuntu13), and I was shocked to get more than 20.000 compiler warnings. So I just started to clean some of them.

I plan to add more compiler warning clean commits if this little commits are accepted and merged.

### DIFF
--- a/Engine/lib/collada/src/dae/daeStringTable.cpp
+++ b/Engine/lib/collada/src/dae/daeStringTable.cpp
@@ -58,7 +58,7 @@ void daeStringTable::clear()
 {
 	unsigned int i;
 	for (i=0;i<_stringBuffersList.getCount();i++)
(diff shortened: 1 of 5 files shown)


eightyeight:
This is definitely significant and something I want to deal with (see #1150). Just quickly, does the warning go away if you use `#if defined(_MSC_VER) && _MSC_VER > 1000`? Would be preferable to having nested `#if`s, but if not then I guess we'll go this way.

raistlin:
Added some more fixes.

eightyeight:
Looks great. I'll give this a quick test later but it should be good to go.

eightyeight:
Thanks for the contribution!

PR #1339, opened 2015-07-02
Remove demo and trial checks

Areloch:
Removing stuff pertaining to demo modes and trials, as they're pointless.


dottools:
Yes! it's been a long time coming to finally get rid of that demoware checking code.

eightyeight:
Oh wow this was a big chunk of code. Thanks for that! I think the one argument for leaving it in is that someone may actually want to use it for their own demos...

...but too bad.

Areloch:
Well, outside the script stuff, the engine-side demo stuff actually isn't even functional. it relied on a demo/ directory for all that stuff that isn't a thing now.

So too bad indeed :P

eightyeight:
Ah. Okay!

PR #1341, opened 2015-07-02
Fixed warning.

raistlin:
Sort struct initialization to match fields declaration avoids a warnings on GCC.

This is a bit tricky, because with this little change I had removed a good amount of warnings, console.h were included almost anywhere.
I have run some demo and all seem to work fine.


eightyeight:
See #1150. Thanks for this!

PR #1342, opened 2015-07-03
offsetof is actually a standard thing nowadays it would seem

Azaezel:
First reported https://www.reddit.com/r/gamedev/comments/3bv7oq/torque_3d_37_released_linux_opengl_client/cspvouc

@dottools found the following:
http://lxr.free-electrons.com/source/include/linux/stddef.h#L15 
http://lxr.free-electrons.com/source/include/linux/compiler-gcc4.h#L14

MSDN:
https://msdn.microsoft.com/en-us/library/dz4y9b9a.aspx

### DIFF
--- a/Engine/source/console/consoleTypes.h
+++ b/Engine/source/console/consoleTypes.h
@@ -48,11 +48,7 @@
 /// @{
 
 #ifndef Offset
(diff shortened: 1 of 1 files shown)


eightyeight:
Oh, awesome! Nice work the both of you.

PR #1343, opened 2015-07-03
Fixed some random gcc's -Wreorder warnings

raistlin:
Continue to work fixing #1150


eightyeight:
Fantastic, thanks for this. Looks good, will give it a quick test then we should be fine.

Issue #1174, opened 2015-02-09
pool->flushWorkItems(); may not work as expected


lukaspj:
I was doing some multithreading on the particlesystem, the code can be found [here](https://gist.github.com/lukaspj/448c8910951ad6f4c232#file-particlethreading-cpp-L3-L11).

But I encountered an issue when there were several threads running, adding some debug info, I got this in the console:
```
Created queue
Finished a Particle Worker Thread
Finished a Particle Worker Thread
Finished a Particle Worker Thread
Finished a Particle Worker Thread
Deleting queue
Deleted queue
Created queue
Finished a Particle Worker Thread
Finished a Particle Worker Thread
Finished a Particle Worker Thread
Deleting queue
First-chance exception at 0x10E818EC (Empty_DEBUG DLL.dll) in Empty_DEBUG.exe: 0xC0000005: Access violation reading location 0xFEEEFEF6.
Deleted queue
```

As you can see, in the last pass, it deletes the queue before all the threads have finished, afaik `pool->flushWorkItems();` should ensure that all the threads have finished.

This is the main code that triggers this issue:
```c++
   OutputDebugStringA("Created queue\n");
   ParticleQueue* queue = new ParticleQueue(mParticlePool->GetParticleHead()->next);
   ThreadPool* pool = &ThreadPool::GLOBAL();

   F32 t = F32(ms) / 1000.0;

   for (U32 i = 0; i < 4; ++i)
   {
      ThreadSafeRef< ParticleUpdateItem > item(new ParticleUpdateItem(queue, t, this, BHVs));
      pool->queueWorkItem(item);
   }

   pool->flushWorkItems();

   OutputDebugStringA("Deleting queue\n");
   delete queue;
   OutputDebugStringA("Deleted queue\n");
```

LuisAntonRebollo:
If i remember correctly ```flushWorkItems``` wait until all jobs are outside workItem queue.... but dont know about working/finished.

You can use ```ThreadSafeRef<ParticleQueue>``` stored on each ```ParticleUpdateItem```.

lukaspj:
Ah I thought that outside workItem queue meant that they had been finished, but I guess it just means that they have been taken care of -or- "initiated"

lukaspj:
Yeah that was simply it, sorry for bothering you, but this whole thing is not exactly perfectly described :)

eightyeight:
[...]

PR #1346, opened 2015-07-05
Allow using ThreadPool synchronously

eightyeight:
Previously, a test was using `ThreadPool::flushWorkItems` as if it would wait for all work items to complete before returning. This is incorrect - see #1174. Actually, `flushWorkItems` just waits until all items have _begun_, so if you have long-running items, they may be still happening.

This PR adds two things:

 1. A method `hasExecuted` on `WorkItem` to see if the item has finished executing.
 2. A method `waitForAllItems` on `ThreadPool` to block until no items are currently executing.

This lets you use the `ThreadPool` completely synchronously, as demonstrated by the new tests.


rextimmy:
Ooh great work :) i had this problem when i was playing around with the thread pools and i ended up testing with intel tbb instead.

eightyeight:
Huh, TBB does look cool.

rextimmy:
Oh i meant i had the synchronizing problems when i was testing out a few things with T3D's ThreadPool. Yep TBB is a very nice library.

Areloch:
I'll be honest. I'm not sure how to test this. Glancing at the changes, I see it adds the stuff, but I don't know how to kick off the testing of it?

eightyeight:
You could run the unit tests ;). `torque.exe runTests.cs`

JeffProgrammer:
I kind of makes sense on how flush originally worked as intended. at least in OpenGL, a flush command simply "flushes" the GL commands to the gpu, while swapBuffers actually is the blocking call. I kinda always assumed that a flush ment not blocking, rather just push everything out of the queue. However, I do like having the option for sync! :+1:

eightyeight:
Thanks, that makes sense!

Issue #1347, opened 2015-07-05
Linux Build not displaying window


avarisclari:
I followed the instructions on the compile page to make the Engine, and when I try running the file it only outputs text in the cli with no window. I can't kill it with ctrl c either, and have to kill it from another interface.

```
Initializing platform...
Input Init:
% 
DebugDrawer Enabled!
GFX Init:
   Null device found
% 
Missing file: core/parseArgs.cs!
main.cs (87): Unable to find function defaultParseArgs
--------- Loading DIRS ---------
Missing file: levels/main.cs!
Error: Unable to find specified directory: levels
Missing file: art/main.cs!
Error: Unable to find specified directory: art
Missing file: scripts/main.cs!
Error: Unable to find specified directory: scripts
Missing file: tools/main.cs!
Error: Unable to find specified directory: tools
% 
--------- Parsing Arguments ---------
Engine initialized...
main.cs (261): Unable to find object: 'Canvas' attempting to call function 'showWindow'
Started up in 0.66 seconds...
% 
```

eightyeight:
What's in the folder adjacent to the executable and `main.cs`? Sounds like you don't have any template files, but you're using `main.cs` from a template. Have a go following [these instructions](http://wiki.torque3d.org/introduction:creating-a-project-without-the-project-manager) and then see if it runs!

avarisclari:
Thanks, that's an improvement, but now it segfaults. 
`````
Binding server port to default IP
UDP initialized on port 0
The requested renderer, D3D9, doesn't seem to be available. Trying the default, OpenGL.
The OpenGL renderer doesn't seem to be available. Trying the GFXNulDevice.
Attempting to create GFX device: GFX Null Device []
Initializing GFXCardProfiler (GFX Null Device Renderer)
   o Chipset : 'NULL Device'
   o Card    : 'GFX Null Device Card'
   o Version : '0'
   o VRAM    : 0 MB
   - Scanning card capabilities...
   - Loading card profiles...
      - No card profile core/profile/GFXNullDeviceRenderer.cs exists
      - No card profile core/profile/GFXNullDeviceRenderer.NULLDevice.cs exists
      - No card profile core/profile/GFXNullDeviceRenderer.NULLDevice.GFXNullDeviceCard.cs exists
      - No card profile core/profile/GFXNullDeviceRenderer.NULLDevice.GFXNullDeviceCard.0.cs exists
Device created, setting adapter and enumerating modes
[...]

PR #1348, opened 2015-07-07
Fix running on Linux / Intel Mesa drivers

hpvb:
This allows Torque3D to run on the open source Linux 3D drivers. There are still many issues but with this small patch it at least starts and is semi-functional.

* Explicitly request a GL 3.2 context
* Enable 'experimental' extensions on glew (This appears necessary to get glew to work with explictly set GL versions)

### DIFF
--- a/Engine/source/gfx/gl/tGL/tGL.h
+++ b/Engine/source/gfx/gl/tGL/tGL.h
@@ -24,8 +24,13 @@
 #define T_GL_H
 #include "GL/glew.h"
 
-// Slower but reliably detects extensions
+#if defined (TORQUE_OS_WIN)
+// This doesn't work on Mesa drivers.
+#define gglHasExtension(EXTENSION) GLEW_##EXTENSION
+#else
+// Slower but reliably detects extensions on Mesa.
 #define gglHasExtension(EXTENSION) glewGetExtension("GL_" # EXTENSION)
+#endif
 
 #endif
 
--- a/Engine/source/gfx/gl/tGL/tGL.h
+++ b/Engine/source/gfx/gl/tGL/tGL.h
@@ -25,7 +25,7 @@
 #include "GL/glew.h"
 
 // Slower but reliably detects extensions
-#define gglHasExtension(EXTENSION) glewGetExtension("GL_##EXTENSION")
+#define gglHasExtension(EXTENSION) glewGetExtension("GL_" # EXTENSION)
 
 #endif
 
--- a/Engine/source/gfx/gl/tGL/tGL.h
+++ b/Engine/source/gfx/gl/tGL/tGL.h
@@ -24,7 +24,8 @@
 #define T_GL_H
 #include "GL/glew.h"
 
-#define gglHasExtension(EXTENSION) GLEW_##EXTENSION
+// Slower but reliably detects extensions
+#define gglHasExtension(EXTENSION) glewGetExtension("GL_##EXTENSION")
 
 #endif
 
--- a/Engine/source/gfx/gl/sdl/gfxGLDevice.sdl.cpp
+++ b/Engine/source/gfx/gl/sdl/gfxGLDevice.sdl.cpp
@@ -83,6 +83,10 @@ void GFXGLDevice::enumerateAdapters( Vector<GFXAdapter*> &adapterList )
     );
 
    SDL_ClearError();
+   SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
+   SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 2);
+   SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
+
    SDL_GLContext tempContext = SDL_GL_CreateContext( tempWindow );
    if( !tempContext )
    {
--- a/Engine/source/gfx/gl/tGL/tGL.cpp
+++ b/Engine/source/gfx/gl/tGL/tGL.cpp
@@ -29,6 +29,7 @@ namespace GL
 {
    void gglPerformBinds()
    {
+      glewExperimental = GL_TRUE;
       GLenum err = glewInit();
       AssertFatal(GLEW_OK == err, avar("Error: %s\n", glewGetErrorString(err)) );
    }
--- a/Engine/source/platformSDL/sdlPlatformGL.cpp
+++ b/Engine/source/platformSDL/sdlPlatformGL.cpp
@@ -13,18 +13,16 @@ namespace PlatformGL
            return;
 
        inited = true;
-       const U32 majorOGL = 4;
+       const U32 majorOGL = 3;
        const U32 minorOGL = 2;
        U32 debugFlag = 0;
 #ifdef TORQUE_DEBUG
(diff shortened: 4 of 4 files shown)


eightyeight:
Ooh, thanks. Do you anticipate any compatibility issues resulting from this? I think if @LuisAntonRebollo's happy with this it can go in!

hpvb:
I don't think so. The original code had the context creation code already in it, but commented out for the game context (there is also a detection context) with a note that it breaks glew. The addition of the enabling of the experimental flag should fix that issue.

From what I understand (which is very little, admittedly) this should not cause problems for cards that were already working.

hpvb:
As an aside: I said 'semi-functional' but it appears that apart from overbright HDR and bloom and some minor graphics glitches on the skydome everything seems to be pretty much ok.

Azaezel:
no off-target side effects noted this end at least.

JeffProgrammer:
I think GLEW needs to be replaced, and here's why:

from what I'm reading, the glewExperimental flag will force GLEW to load entry points even if the extension isn't present. In a core opengl context, you have to query each extension by using glGetStringi() and GLEW was never refactored for the modern standard, it still checks extensions using the deprecated glGetString(GL_EXTENSIONS), which will cause a opengl error (which is why your required to use the glewExperiemental flag). 

More information: http://stackoverflow.com/questions/20822087/why-does-this-crash-when-using-opengl-core-profile

If I am wrong on that GLEW will force load the extension even if its not presesnt, correct me, but that's what I'm seeing.

hpvb:
I was not aware that glew experimental just forced all extensions but I did notice some odd errors related to extensions that my driver didn't implement. If this is indeed what happens replacing GLEW seems like a good idea.

hpvb:
I wrote some testcases against GL3.3 Core profile and I have to agree with your observations. GLEW doesn't even properly detect glGenVertexArrays() which is very much part of 3.3 core (and works if experimental is enabled). With experimental enabled it finds valid entry points for features like GL_EXT_gpu_shader4 for instance which are exposed by mesa libGL but don't have implementations in all drivers.

Maybe GLFW?

hpvb:
I added another small commit to this PR, I don't think this is ready for merging yet but it's something worth discussing maybe.

Engine/source/gfx/gl/tGL/tGL.h
```diff
@@ -24,7 +24,8 @@
 #define T_GL_H
 #include "GL/glew.h"
 
-#define gglHasExtension(EXTENSION) GLEW_##EXTENSION
+// Slower but reliably detects extensions
+#define gglHasExtension(EXTENSION) glewGetExtension("GL_" # EXTENSION)
```
Azaezel: This chunk causes issues on the win8.1/non-SDL side.

hpvb: I think this is because we create GL contexts with different parameters in the win32-native bits of the GL driver. I don't have the hardware to debug this myself though. I can see if I can debug this on a VM maybe. Does Windows come with a software GL renderer that supports GL3+?

hpvb:
This seems to also fix  #1347 (Running on Mesa Radeon drivers).

JeffProgrammer:
How about just implementing extension checking yourself instead of relying on GLEW? I'm sure you could just write an inline function for it.

Engine/source/gfx/gl/tGL/tGL.h
JeffProgrammer: Microsoft's "driver stub" is OpenGL 1.4 (Win XP and below is 1.1) and converts calls to D3D (or software) IIRC

hpvb:
@JeffProgrammer we could possible lift it from here : https://github.com/jamesu/Torque3D/tree/cortexgl2 This fork seems to have (mostly) done away with glew already.

JeffProgrammer:
@hpvb that branch is OpenGL 2.1 based [in case you didn't notice] but yes some stuff could be taken as long as the Steering comittee is willing to slap on the Mode 7 Games and JamesU's copyright notices. :) :) :)

hpvb:
@JeffProgrammer I had in fact not noticed that. If we want to keep the separate win32/ backend of glx/gl I guess the best course of action would indeed be to just get rid of glew and use a different bindings generator and our own extension detection. 

I'll have a look, this is all kind of new to me but I'd really like to get this to work properly. If someone has a suggestion please do tell :)

Engine/source/gfx/gl/tGL/tGL.h
Azaezel: Not exactly a solution I'm in love with, but 
#if defined (TORQUE_OS_WIN)
#define gglHasExtension(EXTENSION) GLEW_##EXTENSION
#else
// Slower but reliably detects extensions
#define gglHasExtension(EXTENSION) glewGetExtension("GL_" # EXTENSION)
#endif

seems to work :/

hpvb:
I have talked to the people in #dri-devel and they recommend using https://github.com/anholt/libepoxy instead of glew. I'm looking into the API to see how much work it would be to use it. When is 3.8 scheduled for release?

Azaezel:
Personal advice just from one of the folks that whines about instability would be slate the overhaul for 3.9, since that'll touch on a lot of stuff that'll need re-testing, and the 3.7 release was left in a state where opengl wasn't left in a funcitonal state for anybody. But as always, I Just 'Work' Here (tm)

Areloch:
We're trying for the first test build for 3.8 to be ready in the next week. So I'd be in agreement with Az if replacing glew is going to be a bit more broad-reaching of a change.

We can roll this in for 3.8, so everything at least works for people, and then you can start looking into what using libepoxy would take so we can look at using that in 3.9 as a long-term solution.

hpvb:
OK, I'll update the pull request with the @Azaezel's ifdef. For 3.8 that'll have to do. Can we document somewhere that postprocessing doesn't work on Mesa? Would a quick fix to detect the mesa renderer and disable postfx for 3.8 be acceptable?

Areloch:
Yeah, I think that'll work for 3.8. 

Running with a few missing elements is better than not running at all.

I can make sure that there's a note about that in the 3.8 release notes.

Azaezel:
Turns out this'll need an alternate detection schema. Plays up on Ubuntu with nvidia drivers. Believe @hpvb mentioned researching an environment variable check specifically for mesa.

Lopuska:
I'm reviewing this. Thank you for this PR and all people contributed on it.
Well, this looks quite important to make next release works also on Mesa drivers without modifying code and @Azaezel added ifdef guards to avoid this interfering on windows. So it's a safe and secure addition.
If someone would like to vote libepoxy as glew replacement we can talk about it through forum. If there is a true interest for the community on use this new extension manager, I can take some time to integrate it quickly and see how it works.

Thank you!

hpvb:
Hi,

I still think the long-term solution should be to move to epoxy and SDL2 for all GL platforms, but in the interim this seems like a workable solution.

Lopuska:
Thanks @hpvb for your opinion on that.
So, let's just make a reminder for epoxy!

Azaezel:
Afraid this one's blowing up ubuntu bad once it hits either:
https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/game/core/scripts/client/postFx/hdr.cs#L408

or 

https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/game/core/scripts/client/postFx/hdr.cs#L392

looking into better solutions than full-on reversion, but could take a bit.

Issue #1349, opened 2015-07-10
Profiler does not work correctly on Linux or Win64


GuyAllard:
On linux and 64bit windows, the profiler timing is incorrect. 
Profiles on these platforms will contain percentage time values < 0 and > 100%. This can be seen in the [enterprise profiler dump](http://torque3d.wdfiles.com/local--files/communityproject%3Aperformance%3Aprofiling-and-optimisation/profilerDumpToFile127904.txt) which was generated on 64 bit windows.

The reason is that the profiler timing ASM (in profiler.cpp) is only compiled when TORQUE_SUPPORTS_VC_INLINE_X86_ASM or TORQUE_SUPPORTS_GCC_INLINE_X86_ASM are defined.

TORQUE_SUPPORTS_VC_INLINE_X86_ASM and TORQUE_SUPPORTS_GCC_INLINE_X86_ASM are not defined for  LINUX or WIN64.

eightyeight:
Ahhhhh that explains it. Wow, okay. Should we consider moving this out of ASM?

eightyeight:
Keeping this open as a reminder to sort the high-res timer out.

JeffProgrammer:
add an option to use the SDL high performance counter. works great and is cross platform.

Areloch:
Moving to 3.9 for now. We'll shift to the SDL counter when using SDL then.

JeffProgrammer:
:+1:

PR #1350, opened 2015-07-11
fills in profiler timer fallback

Azaezel:
as per remmed out lines in other samples. see https://github.com/GarageGames/Torque3D/issues/1349 for report, and https://gist.github.com/Azaezel/2aafb88cf0d642418051 for result

### DIFF
--- a/Engine/source/platform/profiler.cpp
+++ b/Engine/source/platform/profiler.cpp
@@ -155,11 +155,13 @@ U32 endHighResolutionTimer(U32 time[2])  {
 
 void startHighResolutionTimer(U32 time[2])
 {
+   time[0] = Platform::getRealMilliseconds();
 }
 
 U32 endHighResolutionTimer(U32 time[2])
 {
-   return 1;
+   U32 ticks = Platform::getRealMilliseconds() - time[0];
+   return ticks;
 }
 
 #endif


eightyeight:
It's not exactly high-resolution bit it's better than nothing! Though, I wonder if we should look at [this C++11 solution](http://stackoverflow.com/a/21856299/945863). Oh, but of _course_ VS2012 doesn't implement it correctly. [Workaround.](http://stackoverflow.com/a/16299576/945863)

JeffProgrammer:
Why not just use each platform's specific high performance counter? Windows: QueryPerformanceCounter/Frequency, Linux/unix: clock_gettime time.h, OSX: mach timer

That way it stays with the C++03 standard.

dottools:
`std::chrono::high_resolution_clock` is really bad and should not be used. For one you don't know if it's really an alias to just `std::chrono::system_clock`, `std::chrono::steady_clock`, or something else as explained on [CPP Reference](http://en.cppreference.com/w/cpp/chrono/high_resolution_clock). Second neither of the first two clock sources are suitable for benchmarking/profiling code. You want a timer source that only increments on thread CPU usage, which there is not option to choose from in C++11.

Similar to what Jeff H. has said. We should be implementing platform specific timers that are known to be correct or good enough timer sources for profiling.
* For Linux (and Android) that's `clock_gettime(CLOCK_THREAD_CPUTIME_ID, &ts)`, [example](http://www.guyrutenberg.com/2007/09/22/profiling-code-using-clock_gettime/).
* For Windows that's `GetThreadTimes` and combining lpKernelTime and lpUserTime values: [GetThreadTimes() documentation](https://msdn.microsoft.com/en-us/library/ms683237%28VS.85%29.aspx).
* No idea about Mac OS X.

eightyeight:
Roger that @dottools, thanks for the info. @Azaezel I'm going to merge this for now until we can work on a longer-term solution. I'm happy to take that on, unless you need it today ;).

PR #1351, opened 2015-07-13
PVS Cleanup: Repeated expression optimization

Areloch:
Issue found with PVS-Studio:

Many instances of a function or expression being used repeatedly, which can lower performance.

Fixed it in these cases by creating on local var, reference or pointer that's used instead.


Areloch:
My local repo history has gone full idiot on this. I'll have to redo this PR.

PR #1353, opened 2015-07-14
Utilize ++iter rather than iter++ to improve iterator performance.

Areloch:
Issue found with PVS-Studio:

Many places utilize post-incrementation with iterators, but it's better performance to use pre-incrementation.

Resolved by changing the iter++ instances to ++iter;

### DIFF
--- a/Engine/source/console/engineDoc.cpp
+++ b/Engine/source/console/engineDoc.cpp
@@ -706,7 +706,7 @@ static bool dumpEngineDocs( const char *outputFile )
    // Dump pre-declarations for any groups we encountered
    // so that we don't have to explicitly define them.
    HashTable<String,U32>::Iterator iter = smDocGroups.begin();
-   for ( ; iter != smDocGroups.end(); iter++ )
+   for (; iter != smDocGroups.end(); ++iter)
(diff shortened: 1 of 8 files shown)


eightyeight:
I am so saddened that the compiler cannot optimise this. Just sad.

andr3wmac:
@eightyeight : What makes you think the compiler won't optimize it? A lot of these corrections are a) don't rely on the compiler to optimize it for you and b) it does no harm, so may as well correct it to shut PVS Studio up.

Areloch:
Yeah, quite a few issues PVS notes are things that are *probably* correctly handled by the compiler, but are not explicit and therefore there is room for the compiler to handle it poorly.

Changes like this, or moving away from c-style casting tell the compiler *exactly* what's intended, and ensures the outcome. So it's more about consistently ensuring the desired outcome, than "the compiler never gets this right".

Even so, generally good practices to do, so hey ;)

eightyeight:
@andr3wmac whoops, you're right, these should be optimised by the compiler as they're primitive types. Whenever I've seen a question on this subject, the advice has always been 'use pre increment anyway', which translates in my head as 'post increment will not optimise' :confused:

But yeah, I'm not disagreeing with this change.

PR #1354, opened 2015-07-14
Cleanup of ease functions operations.

Areloch:
Issue found with PVS-Studio:

Many instances of a variable being modified inside of a complex operation. This could lead to unintended results depending on the whims of the compiler.

Resolved by restructuring the functions to modify the variable first, then do the rest of the operation.

Also some minor reformatting to be more code-style compliant.

### DIFF
--- a/Engine/source/math/mEase.h
+++ b/Engine/source/math/mEase.h
@@ -110,7 +110,8 @@ class EaseF : public Ease
 
 // simple linear tweening - no easing
 // t: current time, b: beginning value, c: change in value, d: duration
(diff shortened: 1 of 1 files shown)


Areloch:
Btw, best way I found to test this is to load a mission and select the level info. it's ambient light setting uses an ease curve. Open that editor up, and go through the various curve options(I took screencaps of each).

Then implement the fixes, and go back and compare them to the prior results. Everything matched up to the unmodified to me, and I believe Az confirmed as well.

eightyeight:
Oh cool, I was wondering where I might find an ease. Will give it a go soon :)

PR #1356, opened 2015-07-17
Convert un-modified function arguments to const references.

Areloch:
Issue found with PVS-Studio:

A lot of instances where some function args are not actually modified in any way, meaning that it is better for performance to convert them into const references. This prevents an additional copy, which can help performance.

### DIFF
--- a/Engine/source/navigation/torqueRecast.h
+++ b/Engine/source/navigation/torqueRecast.h
@@ -28,10 +28,10 @@
 #include "math/mBox.h"
 
 inline Point3F DTStoRC(F32 x, F32 y, F32 z) { return Point3F(x, z, -y); }
-inline Point3F DTStoRC(const Point3F& point)       { return Point3F(point.x, point.z, -point.y); }
+inline Point3F DTStoRC(const Point3F& point){ return Point3F(point.x, point.z, -point.y); }
 inline Point3F RCtoDTS(const F32* xyz)      { return Point3F(xyz[0], -xyz[2], xyz[1]); }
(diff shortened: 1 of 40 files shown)


Engine/source/navigation/torqueRecast.h
```diff
@@ -28,11 +28,11 @@
...
 
 inline Point3F DTStoRC(F32 x, F32 y, F32 z) { return Point3F(x, z, -y); }
-inline Point3F DTStoRC(Point3F point)       { return Point3F(point.x, point.z, -point.y); }
+inline Point3F DTStoRC(const Point3F& point)       { return Point3F(point.x, point.z, -point.y); }
 inline Point3F RCtoDTS(const F32* xyz)      { return Point3F(xyz[0], -xyz[2], xyz[1]); }
 inline Point3F RCtoDTS(F32 x, F32 y, F32 z) { return Point3F(x, -z, y); }
-inline Point3F RCtoDTS(Point3F point)       { return Point3F(point.x, -point.z, point.y); }
-inline Box3F DTStoRC(Box3F box)
+inline Point3F RCtoDTS(const Point3F& point)       { return Point3F(point.x, -point.z, point.y); }
+inline Box3F DTStoRC(const Box3F& box)
```
eightyeight: Please fix spacing here :)

Areloch:
Fix spaces like this, you mean? I'm assuming lining the rows back up is what you were referring to.

eightyeight:
Perfect, thank you kindly. I'm pretty okay with this PR, though it begs the question whether these functions are *supposed* to not modify their arguments, or if they just *don't*.

You know what. Dumb question. Will test today.

PR #1357, opened 2015-07-17
followobject position caching

Azaezel:
re-use the position of an object we're following (within reason) so we're not constantly regenning a path.


eightyeight:
`lastPos` should probably be initialised to `Point3F::max` or something

Azaezel:
not really. It's testing the distance between the selected object's current position and it's last target position. even on the off chance something popped up in the exact same spot, it'd still go to the same place, after all.

eightyeight:
Hmm, fair enough. Please to initialise it in the `FollowData` constructor though. I know it's not an issue but it makes me itchy.

Azaezel:
K.

eightyeight:
Much appreciated

PR #1358, opened 2015-07-17
reduces rotation transmission size 

Azaezel:
from 64 bits per turret to 14 similar to player. they add up.


eightyeight:
Looks fine to me!

PR #1361, opened 2015-07-17
adds minimum displacement check prior to convexSweepTest to avoid NaNs

Azaezel:
Should note depending on bullet library version, this one may or may not be needed.


eightyeight:
Do the NaNs show up only when 0s are passed? Or is it really small-magnitude vectors that are the issue?

Azaezel:
0, but floating point happens. Best to be a bit paranoid with potential ap fatalities.

eightyeight:
Sure, I just worry that an arbitrary constant might become nonsensical if the world is at a different scale for example. I guess, yeah, even near-zero values might cause issues? Doesn't look like there's any existing constant we can tie this to so it's not a magic number, either :/.

PR #1362, opened 2015-07-17
Removes fatal assertion on duplicated object collisions (meshroads, primarily)

Azaezel:
Removes fatal assertion on duplicated object collisions (meshroads, primarily)


eightyeight:
Thanks!

PR #1363, opened 2015-07-17
allowplayerstep lets folks run up sharper angles than normal.

Azaezel:
That should be off by default.


eightyeight:
I guess that makes sense? I'd complain about breaking defaults but actually the editors probably save out the current settings, so it's not like anyone's game will suddenly start behaving differently.

PR #1365, opened 2015-07-22
Removing pointless null-pointer tests for objects created with new

Areloch:
Issue found with PVS-Studio:

Many instances where we would create a object via a new call, and then check that it was non-null.

This is redundant, as if we ever were in a situation where new failed, we'd be crashing left and right already, so the additional check is wasted processing.


eightyeight:
[More info](http://stackoverflow.com/questions/550451/will-new-return-null-in-any-case).

PR #1369, opened 2015-07-23
Fix NaNs in Collada files

eightyeight:
Thanks to [Jeff Yaskus's cleanup script](http://www.garagegames.com/community/blogs/view/22992).


Areloch:
This looks like it does fix most of the issues with the stock files, though I did note it still had to correct the file paths for the textures/materials. I guess it doesn't like the flat file name, no path setup?

eightyeight:
Hmm, I'm not sure. Merging this for now.

PR #1371, opened 2015-07-23
Random VS warnings

eightyeight:
There were causing a _lot_ of warnings in files that include them.


Engine/source/math/mMatrix.h
```diff
@@ -572,10 +572,10 @@ inline MatrixF operator * ( const MatrixF &m1, const MatrixF &m2 )
    return temp;
 }
 
-inline MatrixF& MatrixF::operator *= ( const MatrixF &m )
+inline MatrixF& MatrixF::operator *= ( const MatrixF &m1 )
```
eightyeight: This is `m1` to align with the previous function above.

PR #1372, opened 2015-07-23
Local variables shadowing other locals

crabmusket:
This causes a lot of warnings in VS. My approach to resolving these has been:
- Can the previous variable be used without being redeclared?
- Can either the shadowing or shadowed variable have a more specific/sensible name?
- Should both variables be renamed?

I don't think I've ever reduced the scope of a variable, though that might have been a viable strategy in some of these cases. I don't anticipate issues - I think I've been fairly careful. But scope is one of those tricky things...

### DIFF
--- a/Engine/source/T3D/accumulationVolume.cpp
+++ b/Engine/source/T3D/accumulationVolume.cpp
@@ -169,7 +169,7 @@ void AccumulationVolume::buildSilhouette( const SceneCameraState& cameraState, V
    // depending on whether we project orthogonally or in perspective.
 
    TempAlloc< U32 > indices( mPolyhedron.getNumPoints() );
-   U32 numPoints;
+   U32 numSilhouettePoints;
 
    if( cameraState.getFrustum().isOrtho() )
    {
@@ -181,7 +181,7 @@ void AccumulationVolume::buildSilhouette( const SceneCameraState& cameraState, V
       // And extract the silhouette.
 
       SilhouetteExtractorOrtho< PolyhedronType > extractor( mPolyhedron );
-      numPoints = extractor.extractSilhouette( osViewDir, indices, indices.size );
+      numSilhouettePoints = extractor.extractSilhouette( osViewDir, indices, indices.size );
    }
    else
    {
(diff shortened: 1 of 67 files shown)


dottools:
This PR is going to have to be reworked a bit, cause for example all you had to do to resolve serverQuery.cpp's warning was change a single line:
```diff
diff --git a/Engine/source/app/net/serverQuery.cpp b/Engine/source/app/net/serverQuery.cpp
--- a/Engine/source/app/net/serverQuery.cpp
+++ b/Engine/source/app/net/serverQuery.cpp
@@ -1275,7 +1275,7 @@ static void processPingsAndQueries( U32 session, bool schedule )
    if ( !gPingList.size() && !waitingForMaster )
    {
       // Start the query phase:
-      for ( U32 i = 0; i < gQueryList.size() && i < gMaxConcurrentQueries; )
+      for ( i = 0; i < gQueryList.size() && i < gMaxConcurrentQueries; )
       {
          Ping &p = gQueryList[i];
          if ( p.time + gPingTimeout < time )
```

Instead of changing the entire `for()` loop to use another iterator variable name. Which makes me think there's more unnecessary changes just like it.

eightyeight:
Hah, you're right about that. I did exactly what you suggest in a couple of places, but I guess I didn't notice I could have done it there.

Areloch:
Given how many files this hits, and how......interestingly written some of these overlaps are, I want to give this more attention to really parse through the changes just to be sure. In my surface-level testing it seemed fine, but I remember several parts in the code where the intent of it was a crapshoot, and figure this PR should get a double-check. I'm pushing it back for the moment and will make sure it goes in early for 3.9.

eightyeight:
Fair enough!

Areloch:
Given how long this(very unfortunately) sat, it's gotten pretty old and out of date(with a number of conflicts)

@Azaezel opted to take another pass at it from the current devhead, so we'll just close this one and go with his just so we don't complicate it with the conflict resolutions on top of it.

Could still be good to cross-reference against for any problem spots on the newer PR(or any subsequent ones)

PR #1374, opened 2015-07-28
"AL: PSSM Cascade Viz" tool-button

Azaezel:
Adds a debug toggle for the script-exposed global found https://github.com/GarageGames/Torque3D/blob/c152ae86f3f09b3c1d736954b352723d274582f1/Engine/source/lighting/advanced/advancedLightBinManager.cpp#L168


eightyeight:
Thanks!

PR #1376, opened 2015-07-28
From Dušan Jocić: early out of treeview entries to prevent crashes

Azaezel:
"when you change some item or remove, it does not resize the array and in that way it create problem with value"


Azaezel:
Among other things, this was causing issues with regenerating collision meshes via the object editor. fatally.

eightyeight:
Oh, excellent. Thanks for this, both of you!

PR #1378, opened 2015-07-29
Properly testing pointer vars aren't null before using them.

Areloch:
Issue found by PVS Studio:

Several instances where we utilize a pointer variable without properly testing that they aren't null first.

### DIFF
--- a/Engine/source/renderInstance/renderPrePassMgr.cpp
+++ b/Engine/source/renderInstance/renderPrePassMgr.cpp
@@ -202,20 +202,20 @@ void RenderPrePassMgr::addElement( RenderInst *inst )
    if ( isMeshInst || isDecalMeshInst )
       matInst = static_cast<MeshRenderInst*>(inst)->matInst;
 
-   // Skip decals if they don't have normal maps.
-   if (!matInst || isDecalMeshInst && !matInst->hasNormalMap())
-      return;
-
-   // If its a custom material and it refracts... skip it.
-   if (  matInst->isCustomMaterial() &&
-         static_cast<CustomMaterial*>( matInst->getMaterial() )->mRefract )
-      return;
-
-   // Make sure we got a prepass material.
-   if ( matInst )
+   if (matInst)
    {
-      matInst = getPrePassMaterial( matInst );
-      if ( !matInst || !matInst->isValid() )
+      // Skip decals if they don't have normal maps.
+      if (isDecalMeshInst && !matInst->hasNormalMap())
+         return;
+
+      // If its a custom material and it refracts... skip it.
+      if (matInst->isCustomMaterial() &&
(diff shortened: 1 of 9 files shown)


eightyeight:
Something that might have been worth investigating is whether any of these are _functionally_ never NULL, and in those cases whether passing a reference would be more appropriate.

Areloch:
Aye, that'd probably be a good thing for a more in-depth cleanup.

That said, Az tested and noted that this is causing some weirdities with terrain rendering, so consider this NOT safe to merge in currently. I'll do more testing tomorrow and correct whatever's derping out.

Engine/source/renderInstance/renderPrePassMgr.cpp
```diff
@@ -203,12 +203,11 @@ void RenderPrePassMgr::addElement( RenderInst *inst )
...
    // Skip decals if they don't have normal maps.
-   if ( isDecalMeshInst && !matInst->hasNormalMap() )
+   if (!matInst || isDecalMeshInst && !matInst->hasNormalMap())
       return;
 
    // If its a custom material and it refracts... skip it.
-   if (  matInst && 
-         matInst->isCustomMaterial() &&
+   if (  matInst->isCustomMaterial() &&
          static_cast<CustomMaterial*>( matInst->getMaterial() )->mRefract )
```
Azaezel: This block change looks to need more study. It's running afoul of the special snowflake terrain code.

Engine/source/renderInstance/renderPrePassMgr.cpp
```diff
@@ -203,12 +203,11 @@ void RenderPrePassMgr::addElement( RenderInst *inst )
       matInst = static_cast<MeshRenderInst*>(inst)->matInst;
 
    // Skip decals if they don't have normal maps.
-   if ( isDecalMeshInst && !matInst->hasNormalMap() )
+   if (!matInst || isDecalMeshInst && !matInst->hasNormalMap())
```
eightyeight: Check order of operations here?

Azaezel:
suggested revision after a few pokes:

   // Get the material if its a mesh.
   BaseMatInstance* matInst = NULL;
   if ( isMeshInst || isDecalMeshInst )
      matInst = static_cast<MeshRenderInst*>(inst)->matInst;

   if ( matInst )
   {
      // Skip decals if they don't have normal maps.
      if (isDecalMeshInst && !matInst->hasNormalMap())
         return;
      
      // If its a custom material and it refracts... skip it.
      if (matInst->isCustomMaterial() &&
         static_cast<CustomMaterial*>( matInst->getMaterial() )->mRefract )
         return;
      
      // Make sure we got a prepass material.      
      matInst = getPrePassMaterial( matInst );
      if ( !matInst || !matInst->isValid() )
         return;
   }

Areloch:
Alright, PR updated, not seeing any problems on my end.

PR #1379, opened 2015-07-29
 C4189 warning cleanups

Azaezel:
Addresses roughly half of the C4189 warnings though the following methodologies:

1) truly unused vars removed
2) vars leading to remmed out code for debugging remmed in turn.

left out:
vars in macros.

### DIFF
--- a/Engine/source/console/compiledEval.cpp
+++ b/Engine/source/console/compiledEval.cpp
@@ -435,11 +435,12 @@ static void setFieldComponent( SimObject* object, StringTableEntry field, const 
 
 ConsoleValueRef CodeBlock::exec(U32 ip, const char *functionName, Namespace *thisNamespace, U32 argc, ConsoleValueRef *argv, bool noCalls, StringTableEntry packageName, S32 setFrame)
 {
+/*
 #ifdef TORQUE_DEBUG
    U32 stackStart = STR.mStartStackSize;
    U32 consoleStackStart = CSTK.mStackPos;
 #endif
-
+*/
    //Con::printf("CodeBlock::exec(%s,%u)", functionName ? functionName : "??", ip);
 
    static char traceBuffer[1024];
@@ -2244,12 +2245,12 @@ execFinished:
    }
 
    decRefCount();
-
+/*
 #ifdef TORQUE_DEBUG
    //AssertFatal(!(STR.mStartStackSize > stackStart), "String stack not popped enough in script exec");
    //AssertFatal(!(STR.mStartStackSize < stackStart), "String stack popped too much in script exec");
 #endif
-
+*/
    return returnValue;
 }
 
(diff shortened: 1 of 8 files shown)


Engine/source/console/compiledEval.cpp
```diff
@@ -435,11 +435,12 @@ static void setFieldComponent( SimObject* object, StringTableEntry field, const
 
 ConsoleValueRef CodeBlock::exec(U32 ip, const char *functionName, Namespace *thisNamespace, U32 argc, ConsoleValueRef *argv, bool noCalls, StringTableEntry packageName, S32 setFrame)
 {
+/*
 #ifdef TORQUE_DEBUG
    U32 stackStart = STR.mStartStackSize;
    U32 consoleStackStart = CSTK.mStackPos;
 #endif
-
+*/
```
jamesu: Probably should just use another define for this instead of commenting the whole thing out - eew!

Engine/source/ts/tsMesh.cpp
```diff
@@ -238,7 +238,7 @@ void TSMesh::innerRender( TSMaterialList *materials, const TSRenderState &rdata,
       TORQUE_UNUSED(indexed);
       TORQUE_UNUSED(type);
 #endif
-
+*/
```
jamesu: Ditto

Engine/source/gui/3d/guiTSControl.cpp
```diff
@@ -578,14 +578,12 @@ void GuiTSCtrl::onRender(Point2I offset, const RectI &updateRect)
...
       const FovPort *currentFovPort = GFX->getStereoFovPort();
       const MatrixF *eyeTransforms = GFX->getStereoEyeTransforms();
-      const MatrixF *worldEyeTransforms = GFX->getInverseStereoEyeTransforms();
       const Point3F *eyeOffset = GFX->getStereoEyeOffsets();
 
       for (U32 i=0; i<2; i++)
       {
          GFX->activateStereoTarget(i);
          Frustum gfxFrustum = originalFrustum;
-         const F32 frustumDepth = gfxFrustum.getNearDist();
```
jamesu: Might be worth storing the original outside the loop, though that would probably just be a case of premature optimization.

Engine/source/ts/tsMesh.cpp
Azaezel: Tossup there really. as-is, after the redefine of the TORQUE_UNUSED macro, all this does is lead to (void)sizeof(var)

Engine/source/gui/3d/guiTSControl.cpp
Azaezel: What would you preffer be done with https://github.com/GarageGames/Torque3D/blob/6b9e5dd15e67b152500a3ed0e3e58c695d2c959d/Engine/source/gui/3d/guiTSControl.cpp#L621 in that case?

Areloch: Somewhat distressing that it's a overridden local var too.

If we're not changing the frustum in there at all, then I'd agree, we should just be able to push the frustumDepth declaration to just before the loop, and then use it again on line 612 and rename THAT var to like, furstumDepthAdjusted or something.

Seem reasonable?

Other than that, I didn't run into any issues with this. So if we get that part tweaked we should be good to roll it in.

Azaezel: From what I recall of pointing @jamesu at this one when I dropped it off, 'the origional' referred to 
Frustum gfxFrustum = originalFrustum; not depth.

PR #1381, opened 2015-07-29
From Dušan Jocić: additional debugdraw entries. 

Azaezel:
DirectionLine, OutlinedText, Capsule


Lopuska:
Quick report after test:
These are primitive useful additions for debug draw purpose and they are functional.
This PR is safe and it doesn't involve any other engine part.

PR #1385, opened 2015-08-03
-wall leads to incredibly long compile times

Azaezel:
(we're talking 5 minutes vs 80, here)

Alternatively, someone more competent at dealing with cmake could look into string concatenation + a switch to expose it as an option for folks that really really want to know about every possible boo-boo.


eightyeight:
Is it because there are so many warnings? :p What compiler/OS?

Azaezel:
codeblock/linux

eightyeight:
All right, let's stick this in. Wall can be added back in easily enough using CXX_FLAGS, or we could provide a checkbox to do it as well.

PR #1386, opened 2015-08-03
Flips SDL_STATIC on to correct it creating bad symlinks

Azaezel:
Despite the variable name, it does not, in fact create a static library, it would seem.


Azaezel:
hold on this one. windows-side side effects.

eightyeight:
Is this related to the email forward? I thought we fixed this by adding a link-time parameter to tell the engine to link to the local SDL?

Azaezel:
yes. it is related to the email forward. no. we did not fix it.

eightyeight:
Rpath was the parameter I was trying to think of. Different compilers treating it differently?

Azaezel:
Apparently.

PR #1387, opened 2015-08-05
Fixes the menubar functionality when using SDL.

Areloch:
This resolves menu order, cleanup and close/re-open issues, as well as crashes on close.

It also modifies the look slightly to look closer to the windows menubar to keep a cohesive look regardless of platform.


Azaezel:
Verified

eightyeight:
Awesome work!

PR #1388, opened 2015-08-05
SDL mouse wheel speed fix.

Areloch:
Default scroll speed wasn't delta-modified, so scroll gui controls were very slow when scrolled via mouse wheel.. This corrects the issue.


eightyeight:
Should the other platform layers also use the pref?

Areloch:
Probably, yeah. That can be done up as a separate PR, i think.

eightyeight:
Ok. Also should horiz scrolling also be multiplied?

Azaezel:
Frankly, hard to reproduce one way or the other without a custom mouse, but by the same token, couldn't hurt...

Areloch:
Excellent call.

eightyeight:
I'm happy with this

PR #1391, opened 2015-08-07
warning C4706: assignment within conditional expression

Azaezel:


eightyeight:
I'd prefer `GameBase *gb = gc ? gc->get...() : NULL;` to avoid more of #1351. But good idea getting it out of the conditional

Azaezel:
so
       GameBase *gb = gc ? gc->getCameraObject() : NULL;
      if (gb)
?

eightyeight:
You got it. Thoughts?

Areloch:
Yeah, I like that better as well.

eightyeight:
:+1:

PR #1394, opened 2015-08-11
release the mouse from window constraints when poping up a window prompt

Azaezel:
like, say, assertfatal, to allow clicking in the popup.


eightyeight:
Appreciated! Looks okay to me.

Areloch:
So this issue only hit the linux build as far as we know, right?

eightyeight:
Yep

PR #1395, opened 2015-08-12
credit to @MusicMonkey5555 for spotting. asserts for Div/NULLs 

Azaezel:
with multi-element classes

Also includes his magnitude and normalize safe alts


Azaezel:
focused subset of https://github.com/GarageGames/Torque3D/pull/1182 to get that much in at least.

PR #1398, opened 2015-08-13
Update messageBox.ed.cs

Duion:
Must be "Canvas" instead of "0"


Duion:
I just noticed I did it only for the Full template, I forgot that it is duplicated in the Empty template also.

Areloch:
i managed to completely brainfart on that.

I'm doing a companion PR to this to add the change to the empty template as well.

eightyeight:
Thanks for the fix @Duion and and @Areloch!

PR #1401, opened 2015-08-18
Adds a verifyCompatibility method to the Win32FileSystem to report case-sensitivity issues

Azaezel:
-triggered during debug only
-still need to expand that to handle bad directories.

### DIFF
--- a/Engine/source/platformWin32/winVolume.cpp
+++ b/Engine/source/platformWin32/winVolume.cpp
@@ -236,6 +236,14 @@ Win32FileSystem::~Win32FileSystem()
 {
 }
 
+void Win32FileSystem::verifyCompatibility(const Path& _path, WIN32_FIND_DATAW _info)
+{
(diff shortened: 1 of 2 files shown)


Azaezel:
doesn't catch em all, but does catch more than a few. example output:
https://gist.github.com/Azaezel/84b902169a9a66a6735a#file-gistfile1-txt-L630-L810

eightyeight:
I like this idea; thanks for this! Is OSX not also case-insensitive and would benefit from having this check? That can happen in the future though.

Azaezel:
OSX needs more work to be verified functional Possibly up to and including yet another pass at the opengl layer. (and for myself, still need to save up to get one to compile on instead of having to throw builds to my artists). 

Do wish https://msdn.microsoft.com/en-us/library/windows/desktop/aa364419(v=vs.85).aspx didn't explicitly require the operating system itsself to be set up for case sensitivity in order to properly trigger based on FIND_FIRST_EX_CASE_SENSITIVE to catch the directories, but perhaps we'll have better luck once we do get to the apple end there.

PR #1402, opened 2015-08-18
Files caught by https://github.com/GarageGames/Torque3D/pull/1401

Azaezel:


eightyeight:
To be a real nuisance, honestly I would have gone the other way and edited the code to point to lowercase named files. Oh, but most of these cases are auto-named images from the creator, right? So they're named after classes, which have uppercase naming. Ugh :/. Also `cameraBookmark` -> `Camera`? Is the path currently incorrect?

Azaezel:
Going off the icon displays, the camerabookmark and camera thing got flipped because the data is identical, so git just flipped a coin on intent there.

As far as methodologies goes, from @Areloch in his own followup:

https://github.com/GarageGames/Torque3D/blob/development/Engine/source/gui/worldEditor/editorIconRegistry.cpp#L88-L163https://github.com/GarageGames/Torque3D/blob/development/Engine/source/gui/worldEditor/editorIconRegistry.cpp#L136

and it also ducks around about 5 additional spots via script before resolving, so yeah, there's additional vectors that could be explored. Just not necessarily more expedient ones.

PR #1410, opened 2015-09-16
Light animation brightness fix

Areloch:
Corrects light animation to adhere to the brightness level set in the light's params instead of sticking with the default 0-1 brightness range.


Azaezel:
looks good to go.

PR #1413, opened 2015-09-17
partial reversion for #1333

Azaezel:
To be honest, can't remember how I was intending to fix that, but this one's causing it to fail to profile twice in a row, so kill it with fire.

### DIFF
--- a/Engine/source/platform/profiler.cpp
+++ b/Engine/source/platform/profiler.cpp
@@ -214,12 +214,10 @@ Profiler::~Profiler()
 void Profiler::reset()
 {
    mEnabled = false; // in case we're in a profiler call.
-   ProfilerData * head = mProfileList;
-   ProfilerData * curr = NULL;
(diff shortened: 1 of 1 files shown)


Lopuska:
I fixed it in the correct way. Thanks Az!

dottools:
Whitespace galore!
```C++
   while ( curr )
   {
      head = curr->mNextProfilerData;
      free( curr );

      if ( head )
         curr = head;
      else
         curr = NULL;
   }
```
Should look like this:
```C++
   while (curr)
   {
      head = curr->mNextProfilerData;
      free(curr);

      if (head)
         curr = head;
      else
         curr = NULL;
   }
```

Friends shouldn't let friends whitespace all the syntax... :p

Lopuska:
That's true buddy! I'll remove whitespace in a separate PR. Thanks for this. :P

PR #1414, opened 2015-09-18
Gui profiler

ChelaruCatalin:
Add a  Gui to the profiler from t3d.


eightyeight:
This is a great idea!

ChelaruCatalin:
When will i see this in the main branch ?

Azaezel:
Needs cherrypicking out, but can confirm this works. It's how we found the crash resolved in https://github.com/GarageGames/Torque3D/pull/1413

PR #1419, opened 2015-09-22
mQuat.h change to fix QuatF::angleBetween

irei1as:
The old version doesn't have that 2.0f in the return that seems to be needed.
Also added normalizing inside so it can be used directly for not-normalized quaternions too.

### DIFF
--- a/Engine/source/math/mQuat.h
+++ b/Engine/source/math/mQuat.h
@@ -227,7 +227,7 @@ inline F32 QuatF::dot( const QuatF &q ) const
 
 inline F32 QuatF::angleBetween( const QuatF & q )
 {
-   // angle between two quaternions.
+   // angle between two normalized quaternions.
    return mAcos(q.dot(*this)) * 2.0f;
 }
 
--- a/Engine/source/math/mQuat.h
+++ b/Engine/source/math/mQuat.h
@@ -227,8 +227,13 @@ inline F32 QuatF::dot( const QuatF &q ) const
 
 inline F32 QuatF::angleBetween( const QuatF & q )
 {
-   // angle between to quaternions
-   return mAcos(x * q.x + y * q.y + z * q.z + w * q.w);
+   // angle between two quaternions
+   QuatF base(x,y,z,w);
+   base=base.normalize();
+   QuatF q_norm=q;
+   q_norm=q_norm.normalize();
+   return 2.0f*mAcos(base.dot(q_norm));
 }
+
 
 #endif // _MQUAT_H_
--- a/Engine/source/math/mQuat.h
+++ b/Engine/source/math/mQuat.h
@@ -229,11 +229,10 @@ inline F32 QuatF::angleBetween( const QuatF & q )
 {
    // angle between two quaternions
    QuatF base(x,y,z,w);
-   base=base.normalize();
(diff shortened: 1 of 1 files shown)


eightyeight:
I don't know the maths so I can't verify. It would be really good to have some unit tests for this.

irei1as:
Let me first update with a second commit because normalize() doesn't need asignation to change the quaternion object.

Now, some tests:
Inside a .cpp (for example a new one in source):

We need

#include "console/console.h"
#include "math/mMathFn.h"
#include "math/mQuat.h"
#include "console/engineAPI.h"

Or just use mConsoleFunctions.cpp and add only #include "math/mQuat.h".
Now at the end of that file we add this new console function:


DefineConsoleFunction( mAngleBetween, F32, (Point4F a, Point4F b),,
    "Calculate the angle between two quaternions.\n"
    "@param a b two quaternions in Point4F format."
    "@returns The angle."
    "@ingroup Math" )
{
   	QuatF aq(a.x,a.y,a.z,a.w);
	QuatF bq(b.x,b.y,b.z,b.w);
	return aq.angleBetween( bq );
}



That small addition was to be able to use angleBetween() from torquescript.

After compiling, inside any .cs to execute with torquescript we add a function to make quaternions:
(It's the same as QuatF & QuatF::set( const Point3F &axis, F32 angle ) inside mQuat.cpp)

function makequat(%axis, %angle)
{
	%sinHalfAngle=mSin(%angle/2);
	%cosHalfAngle=mCos(%angle/2);
	return (%axis.x*%sinHalfAngle SPC %axis.y*%sinHalfAngle SPC %axis.z*%sinHalfAngle SPC %cosHalfAngle);
}



Now in the console of the game we can test with:

$quat1=makequat("0 0 1",mDegToRad(45));
$quat2=makequat("0 0 1",mDegToRad(90));
mRadToDeg(mAngleBetween($quat1,$quat2));
//we spect 45º between $quat1 and $quat2 and that's what we get (except rounding errors)
//with the old value we had an incorrect 45/2

---------------------
The normalize part is because the dot product ( x·y=|x||y|cos(A) , with normalized |x|=|y|=1).
The use of the 2.0f*Acos() is because in the definition of quaternions what is used is cos(x/2) instead of cos(x).

I don't know that much of these things myself so I'm not sure if there are any issue with weird angles or other kind of normalizations, sorry. But for basic axis and angles this way seems to work.

Lopuska:
I checked the mathematics behind your implementation. It's correct! Thanks so much for your contribution.

Are you sure it's a good idea to normalize the quaternion? What about to give this responsibility to the caller? It's pretty much costly because I see 2 object construction, 1 copy and 2 normalization.
It's important to have our vector math as fast as possible and try to don't wast CPU cycles since we have some area that are already performance critical.
I could suggest this implementation I modified from your code.
What do you think?

return mAcos(q.dot(*this)) * 2.0f;

irei1as:
That's quite true.
Rotation math with quaternions needs normalizing so it's a good chance the one using this has them already normalized for other uses. To force it again hidden inside is kind of a waste. Let me change it like that.

Lopuska:
Thanks @irei1as looks perfect! ;)

PR #1421, opened 2015-09-23
CMake NMake fix to place exe in game directory

blackwc:
This fixes CMake NMake build generation to place the exe in the game directory


eightyeight:
Nice one!

PR #1425, opened 2015-09-25
Templates sfx autodetect fix

blackwc:
%providerA in sfxCompareProvider function is a vector instead of the first field. This fixes that and prevents defaulting to Null SFX Provider.


blackwc:
This is just a temporary hack. Ideally sort needs to be fixed.

eightyeight:
This sounds good, but why did you close it?

PR #1426, opened 2015-09-25
backend correction for https://github.com/GarageGames/Torque3D/pull/1425

Azaezel:


eightyeight:
Looks reasonable at a glance.

Issue #1219, opened 2015-02-22
Mac OSX Terrain does not render in release mode


JeffProgrammer:
Terrain renders fine in debug, but when it compiles in release, terrain just does not render properly at all. It basically becomes a flat sandy plane :/

jnoyola:
It seems to be caused by the difference in optimization level. Adding `__attribute__((optnone))` before `U64 hash64(...)` in core/util/hashFunction.cpp fixed the issue for me (after way too much debugging).

eightyeight:
Strange. Thanks for checking that out! Looking through [the function](https://github.com/GarageGames/Torque3D/blob/development/Engine/source/core/util/hashFunction.cpp#L214) there doesn't seem to be anything too crazy, but then again I wouldn't actually know :p.

jnoyola:
I'll try to take a look at the assembly today. I remember seeing warnings in the function but I didn't have time to look too closely yet. It's tough managing a tech internship, sleep, and a game-design addiction (not to mention occasionally -- read: several hours a day -- playing some games too).

JeffProgrammer:
are you compiling with 03 optimization by chance? 03 isn't really considered "stable" per say, so i hear. Try 02

eightyeight:
Man, I wish I had time to actually play games these days :p.

jnoyola:
-Os (fastest, smallest) is actually the default and what I use.
LLVM seems to know something we don't, as any optimization level above -O0 optimizes away the entire function:
```
pushl  %ebp
movl   %esp, %ebp
popl   %ebp
retl
```

eightyeight:
Well. That's amusing. Which compiler is this, please?

jnoyola:
Apple LLVM 6.1

Azaezel:
Straight up bizarre given the usage-context...:
https://github.com/GarageGames/Torque3D/blob/c152ae86f3f09b3c1d736954b352723d274582f1/Engine/source/shaderGen/shaderGen.cpp#L468-L472

PR #1428, opened 2015-10-01
cuts the shadergen hashkey generator down from a 64 bit to a 32 bit key.

Azaezel:
(MacOSX 32 bit compatibilty errors. was generating 2k+ shader files. may also address  https://github.com/GarageGames/Torque3D/issues/1219)


JeffProgrammer:
Can confirm works on OSX :+1:

PR #1432, opened 2015-10-06
Backend correction for the rigid vs rigid collision resolver:

Azaezel:
First line is to ensure similar behavior to current regarding pushback on the object doing the colliding.
Second line applies an impulse to the rigid that was collided with.

### DIFF
--- a/Engine/source/T3D/rigid.cpp
+++ b/Engine/source/T3D/rigid.cpp
@@ -156,7 +156,7 @@ bool Rigid::resolveCollision(const Point3F& p, const Point3F &normal, Rigid* rig
       return false;
 
    // Compute impulse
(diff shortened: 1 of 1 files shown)


Azaezel:
Special note, this is an extract from the reactiverigids fork, however given the desire to revamp the hierarchy referenced by https://github.com/Azaezel/Torque3D/commit/98a792caabc9a51e3864f42a9969417f21d62170#diff-cae7cf0b1281b14a21f867f1d3e7a7d3R1238 this is purely to ensure no false-positives crop up when converting the system to a more plugin-with-fallbacks archetecture.

Lopuska:
It's correct for me. Thanks Az!

PR #1433, opened 2015-10-06
The TypeCommand type brings up a full notepad-esque interface.

Azaezel:
http://i.imgur.com/fB44GBl.png


Azaezel:
https://gist.github.com/Azaezel/c246aee4dbad00f4f019 quickiedump of additional spots to eyeball if we want to add that functionality to more places

eightyeight:
Great idea.

PR #1434, opened 2015-10-07
fullscreen and windowed mode cli fix

blackwc:
"-fullscreen" and "-windowed" were not working properly. This fixes that for Full and Empty templates.


Azaezel:
For folks getting turned around wonderign where that $WORD::FULLSCREEN is coming from:

https://github.com/GarageGames/Torque3D/blob/c152ae86f3f09b3c1d736954b352723d274582f1/Templates/Empty/game/core/main.cs#L24-L29

Lopuska:
Safe and functional. Thanks for fixing this bug.

PR #1436, opened 2015-10-07
missing ribbon shaders, empty template

Azaezel:


eightyeight:
IIRC these were deliberately left out to make the empty template more empty. Though actually that probably only applies to the art. I'm happy to include these.

PR #1438, opened 2015-10-09
explosion cover miscalc

Azaezel:
%distscale can actually end up negative. causes miscalculations for applyimpulse and the like (or even healing if you've hacked in the capacity for negative damage)


eightyeight:
LGTM

PR #1439, opened 2015-10-12
ensures opengl texSpaceMat is initialized from the get-go 

Azaezel:
cleans up those
0(473) : warning C7050: "objToTangentSpace[0]" might be used before being initialized
0(473) : warning C7050: "objToTangentSpace[1]" might be used before being initialized
0(473) : warning C7050: "objToTangentSpace[2]" might be used before being initialized
warnings.


eightyeight:
Sounds like a good idea!

PR #1441, opened 2015-10-13
(mostly) resolves errors involving beginning queries on the same object in succession without a break.

Azaezel:
reference:
https://www.opengl.org/sdk/docs/man3/xhtml/glBeginQuery.xml
https://www.opengl.org/sdk/docs/man3/xhtml/glIsQuery.xml


Azaezel:
ran into a few more missing bits.

PR #1443, opened 2015-10-14
TAML, Assets and Modules implementation

Areloch:
Initial implementation of the TAML, Asset and Module systems ported from T2D.

This is everything necessary for the base implementation and doesn't override any existing functionality. It comes with a blank example asset and a ShapeAsset that loads a shape file and stores the TSShape for later reference.


Areloch:
Technically TAM was an acronym at that point, but yeah, it did look like a typo, huh ;)

I tweaked the title to be clearler.

eightyeight:
Oh, I see! Sorry :joy:

Azaezel:
notes thusfar:
changes the signature of addProtectedField, so will need to go through and add &defaultProtectedWriteFn, entries, like so:

https://github.com/GarageGames/Torque3D/blob/2044b2691e1a29fa65d1bdd163f0d834995433ce/Engine/source/T3D/fx/particle.cpp#L207-L208

to
   addProtectedField( "sizes", TYPEID< F32 >(), Offset(sizes, ParticleData), &protectedSetSizes, 
      &defaultProtectedGetFn, &defaultProtectedWriteFn, PDC_NUM_KEYS,

there's a fairly full (though incompatible) list https://github.com/lukaspj/Torque3D/commit/fa89f62d022ffd17e8b6c8158da1d9fa14920374

do note that as a function-pointer, new AbstractClassRep::WriteDataNotify() should be replaced where possible with &defaultProtectedWriteFn as in the above example as well.

Azaezel:
Scratch the above, we found a less intrusive fix.

PR #1444, opened 2015-10-16
Removed unused vertex formats from ScatterSky

rextimmy:
I noticed this while working on DX11 stuff, scattersky has unused vertex formats. Thought i would do a separate PR in case it got lost in the DX11 branch

### DIFF
--- a/Engine/source/environment/scatterSky.cpp
+++ b/Engine/source/environment/scatterSky.cpp
@@ -85,8 +85,6 @@ const F32 ScatterSky::smViewerHeight = 1.0f;
 GFXImplementVertexFormat( ScatterSkyVertex )
 {
    addElement( "POSITION", GFXDeclType_Float3 );
-   addElement( "NORMAL", GFXDeclType_Float3 );
-   addElement( "COLOR", GFXDeclType_Color );
 }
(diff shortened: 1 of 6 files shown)


Engine/source/environment/scatterSky.cpp
```diff
@@ -85,8 +85,6 @@ const F32 ScatterSky::smViewerHeight = 1.0f;
 GFXImplementVertexFormat( ScatterSkyVertex )
 {
    addElement( "POSITION", GFXDeclType_Float3 );
-   addElement( "NORMAL", GFXDeclType_Float3 );
-   addElement( "COLOR", GFXDeclType_Color );
 }
```
Azaezel: Good catch. that bein the case, why not kill it entirely, given GFXImplementVertexFormat( GFXVertexP ) :
https://github.com/GarageGames/Torque3D/blob/2044b2691e1a29fa65d1bdd163f0d834995433ce/Engine/source/gfx/gfxVertexTypes.cpp#L27

rextimmy:
yeah good point, i'll fix it up

rextimmy:
Better?

PR #1059, opened 2014-12-16
Volumetric fog   redo

Winterleaf:
Had to redo #1030, I included the opengl stuff.

Someone needs to test the opengl stuff.

PR #1445, opened 2015-10-18
Volumetric Fog with corrections

Areloch:
Based on the Winterleaf PR #1059, this adds corrections so that it works with HDR being active, and also utilizes ambient color so that it works with day/night cycles, rather than being unlit during low-light settings.


eightyeight:
Nice!

rextimmy:
A few of the shaders have text comments from the vehicle scripts. Example https://github.com/GarageGames/Torque3D/pull/1445/files#diff-d60227b88a44f1c0efc02bff45296431R23

Areloch:
Yeah, it looks like there's just several chunks of this that are a bit weird. I'm working through the whole thing manually to try and shore up those weird bits that snuck past everyone the first go around.

Areloch:
Redid this with #1478

PR #1447, opened 2015-10-24
improved radio button

blackwc:
The radio button looks a bit weird so I changed it.


eightyeight:
The bottom half is the disabled states, right? Having that greyed out does look good.

blackwc:
Yes. If you attempt to dot it when it is already dotted it will display the last tile.

I've updated it so that when the radio is disabled it will display at 50% opacity like the buttons.

The reason I say the original looks weird is because the active dot wasn't a full circle which didn't look right to me.

Areloch:
Yeah, I like how you have it now, where it's only the grayed out if you're trying to re-click(to give it a bit of visual feed back).

PR #1454, opened 2015-10-30
adds a dedicated server cmake toggle

Azaezel:


eightyeight:
We already have [something](https://github.com/GarageGames/Torque3D/blob/2044b2691e1a29fa65d1bdd163f0d834995433ce/Tools/CMake/torque3d.cmake#L83) - it should probably be removed in favour of this.

PR #1460, opened 2015-11-10
fillin for fallbacks for filesystem funcs

Azaezel:


eightyeight:
I guess this makes sense!

Azaezel:
Yeah, the docs for https://github.com/GarageGames/Torque3D/blob/2044b2691e1a29fa65d1bdd163f0d834995433ce/Engine/source/console/engineAPI.h#L2591 were wrong. need to feed it both variable fallbacks in multi-param functions. (Do note that does mean there's likely quite a few more with the same misbehavior, if someone wants to go over the full list with a fine toothed comb)

PR #1466, opened 2015-11-15
fix for loading cached dts from zip archives

blackwc:
This fixes a bug where if you attempted to load cached dts files from a zip archive it would fail. This was because it was attempting to extract the file time information from non-existent files outside of the archive. Thus, canLoadCachedDTS would return false.

I added a line to disable the file change notification errors for directories that don't exist or are in zip archive.

Additionally, I added *.pak to the search pattern for archives.

### DIFF
--- a/Engine/source/ts/collada/colladaShapeLoader.cpp
+++ b/Engine/source/ts/collada/colladaShapeLoader.cpp
@@ -514,22 +514,22 @@ bool ColladaShapeLoader::canLoadCachedDTS(const Torque::Path& path)
    Torque::Path cachedPath(path);
    cachedPath.setExtension("cached.dts");
 
-   // Check if a cached DTS newer than this file is available
-   FileTime cachedModifyTime;
-   if (Platform::getFileTimes(cachedPath.getFullPath(), NULL, &cachedModifyTime))
-   {
-      bool forceLoadDAE = Con::getBoolVariable("$collada::forceLoadDAE", false);
+   Torque::FS::FileNode::Attributes a1, a2;
 
-      FileTime daeModifyTime;
-      if (!Platform::getFileTimes(path.getFullPath(), NULL, &daeModifyTime) ||
-         (!forceLoadDAE && (Platform::compareFileTimes(cachedModifyTime, daeModifyTime) >= 0) ))
-      {
-         // DAE not found, or cached DTS is newer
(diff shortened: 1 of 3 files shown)


eightyeight:
This seems like a good idea! This is looking for existing cached DTS files in zips, correct? Cache files don't get created inside zips if there aren't any when the zip is created?

blackwc:
I'll need to add [this](https://github.com/GarageGames/Torque3D/pull/1466/files#diff-0555e026c0fb978ade558c584fe9c81cL521) back in, no doubt.

rasteron:
Hey guys, got interested also in seeing this to work as it completes the process of packaging dts files in zip format. However, after patching a recent build with the files, I did notice that it does not cover the proper reading of the equivalent materials.cs file or particularly the equivalent compiled materials.cs.dso.

Thanks.

Engine/source/ts/collada/colladaShapeLoader.cpp
```diff
@@ -514,21 +514,21 @@ bool ColladaShapeLoader::canLoadCachedDTS(const Torque::Path& path)
    Torque::Path cachedPath(path);
    cachedPath.setExtension("cached.dts");
 
-   // Check if a cached DTS newer than this file is available
-   FileTime cachedModifyTime;
-   if (Platform::getFileTimes(cachedPath.getFullPath(), NULL, &cachedModifyTime))
-   {
-      bool forceLoadDAE = Con::getBoolVariable("$collada::forceLoadDAE", false);
```
Lopuska: Hello! Your PR is good! Thanks for contribution.
You may want to restore this flag as it's used in the engine and as well from the editor to force the load of *.dae instead of cached *.dts.

Like:
if( Con::getBoolVariable("$collada::forceLoadDAE", false) )
   return false;

Then we can merge.
Thanks.

blackwc:
@Lopuska didn't mean to leave that out. All good now.

blackwc:
@rasteron I haven't noticed any of those issues.

Lopuska:
Dear @blackwc,
Me and @Alex-doc have decided to improve this in a different and more proper way:
Check #1540 and feel free to comment and give your opinion on that.

Closing this as we have a more consistent version.
Thanks for your help and contribution.

PR #1469, opened 2015-11-19
Recurse dump directories fix

Areloch:
Fix for the directory scan for modules so it doesn't trim off characters in the path.


dottools:
What is going on with the commits and it looks like you have a bug on line 1391.

blackwc:
@Areloch works for me

Issue #1458, opened 2015-11-07
Bullet shells bounce off water


Duion:
The bullet shells do not sink into water blocks, but stay on top of the surface like the water was a solid wall.

eightyeight:
Obviously bullet shells float!

blackwc:
[relevant](https://github.com/GarageGames/Torque3D/blob/development/Engine/source/T3D/debris.cpp#L44)

Azaezel:
I'd more eyeball the side effects of stripping https://github.com/GarageGames/Torque3D/blob/8253bcc343d682223efc20c58f4c0bd44fd07c5f/Engine/source/environment/waterObject.cpp#L209 on down to it's own type

blackwc:
aslo [relevent](https://github.com/GarageGames/Torque3D/blob/development/Engine/source/T3D/debris.cpp#L762)

PR #1470, opened 2015-11-20
removes StaticObjectType flag from water objects.

Azaezel:
https://github.com/GarageGames/Torque3D/issues/1458


eightyeight:
I'm just waiting to see what other code this will break, haha. If there's nothing obvious it seems fine to me.

Azaezel:
Only thing in particular that looked like it may at some point go hinky was the navmesh gen, and it *seemed* to work out ok anyway. since you wrote that, might be a good notion to re-review that subsection just in case @Areloch or I overlooked a case.

PR #1472, opened 2015-11-22
Fix unknown method vsprintf

Ragora:
This is causing builds to fail on my Linux Mint box. However, I don't really know where to stick this include.

### DIFF
--- a/Engine/source/platform/platform.h
+++ b/Engine/source/platform/platform.h
@@ -23,6 +23,7 @@
 #ifndef _PLATFORM_H_
 #define _PLATFORM_H_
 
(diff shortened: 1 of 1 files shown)


eightyeight:
Thanks! Linux fixes are always appreciated.

Lopuska:
Hello @Ragora!
Thanks for your contribution! We already fixed this by setting custom Torque's vsprintf.
Closed as duplicate PR: #1516

Ragora:
@Lopuska Ah, I was wondering why this was hanging around so long. Thanks!

PR #1474, opened 2015-11-27
 mDirtyTiles changed from std::queue to a vector

Azaezel:
allows us to leverage  .push_back_unique(&foo); and .clear();
(part 1 in a series)


Engine/source/navigation/navMesh.cpp
```diff
@@ -844,7 +844,7 @@ unsigned char *NavMesh::buildTileData(const Tile &tile, TileData &data, U32 &dat
 
    // Check for no geometry.
    if(!data.geom.getVertCount())
-      return false;
+      return NULL;
```
eightyeight: Good catch there

eightyeight:
I'm fine with this!

PR #1475, opened 2015-11-27
NavMeshUpdateAll leak suppression (not 100% preventative)

Azaezel:
1) Reset the build when adding potential dirties to the list to ensure it isn't trying to kill off a dirty entry that it's passed by.
2) mSaveIntermediates = true; causes leakage even with all this. still tracking that end.
3) Need to remove a dead tile whether there's new data to replace it with or not. Empty tiles are an entirely possible case. Even a probable one if you have high verticality in a level.
4) Likewise you don't want to re-feed the same geometry list for a given tile in case the conditions have changed.
5) If no vertcount then clear all geometry entries. (that one might be paranoia)


Engine/source/navigation/navMesh.cpp
```diff
@@ -113,7 +113,11 @@ DefineConsoleFunction(NavMeshUpdateAll, void, (S32 objid, bool remove), (0, fals
    for(U32 i = 0; i < set->size(); i++)
    {
       NavMesh *m = static_cast<NavMesh*>(set->at(i));
-      m->buildTiles(obj->getWorldBox());
+	   if (m)
```
dottools: Uh oh tabs vs spaces :(

Engine/source/navigation/navMesh.cpp
```diff
@@ -113,7 +113,11 @@ DefineConsoleFunction(NavMeshUpdateAll, void, (S32 objid, bool remove), (0, fals
    for(U32 i = 0; i < set->size(); i++)
    {
       NavMesh *m = static_cast<NavMesh*>(set->at(i));
-      m->buildTiles(obj->getWorldBox());
+      if (m)
```
eightyeight: `static_cast` doesn't return NULL the same way `dynamic_cast` does, so this check is redundant! Unless you're worried about `set->at` returning NULL?

eightyeight:
This looks good to me!

Engine/source/navigation/navMesh.cpp
Azaezel: hadn't caught that was static, actually.

Lopuska:
Tested and reviewed. All looks fine!

Azaezel:
For folks that get confused by the removal reversion: the method was discussed as a clearer alternative label to NavMeshUpdateAll so folks knew what it was supposed to do, with the agreement being that NavMeshUpdateAll will be removed come 4.0, but kept for now for backwards compatibility.

Lopuska:
Great @Azaezel! Thanks for this. We should just to remind the removal of this func for 4.0 release (:

PR #1480, opened 2015-12-02
Fix NULL pointer Crashes in WorldEditor::selectObject & unselectObject

Ragora:
Relatively simple fix, the engine will crash because this code fails to return if the input obj is NULL:

https://github.com/GarageGames/Torque3D/blob/development/Engine/source/gui/worldEditor/worldEditor.cpp#L2834

Same for selectObject.
This stack is generated in unselectObject's case: http://pastebin.com/nUzFJN5u

So it pretty much just attempts to make a call against a NULL pointer if you don't give them a valid ID :) (Try EWorldEditor.unselectObject(0))

As an aside, I fixed my line endings so we don't have spastic messes for commits anymore.


eightyeight:
Looks fine to me!

PR #1488, opened 2015-12-21
Physx3 cmake

rextimmy:
Support for PhysX 3.x with CMake. I will make a thread very soon about this, basically NVidia doesn't supply binaries any more for PhysX and only source code access so you have to compile yourself. I will outline the steps in the forum post.

Also had to add two new macros for adding libs with CMake, addLibRelease and addLibDebug. The original addLib macro will work just as before but these two new macros allow the use of different libs for debug/release modes. The original addLib obviously still adds the lib to both builds so this will not change any other current CMake modules. Perhaps I should have split this into two separate PR's??

Tested with latest PhysX 3.3.4

### DIFF
--- a/Tools/CMake/basics.cmake
+++ b/Tools/CMake/basics.cmake
@@ -141,6 +141,8 @@ endmacro()
 ###############################################################################
 ### Linked Library Handling
 ###############################################################################
+
+#addLib will add to both debug and release builds
 macro(addLib libs)
    foreach(lib ${libs})
         # check if we can build it ourselfs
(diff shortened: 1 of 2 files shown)


rextimmy:
Oh yeah if you are wondering why I didn't add VS 2010 CMake support for PhysX 3, that is because NVidia no longer support it.

Areloch:
It's only 5 years old! :x

Seriously though, this is cool. I'll be giving this a whirl along the weekend for sure.

rextimmy:
Ok sweet, i'll install linux tonight and see if i can get it working there too and just amend this PR if i do.

PR #1492, opened 2016-01-04
AiPlayer run animation fix

rasteron:
While trying out AI players, I noticed a weird sliding motion without any run or side animation most of the time. After doing some tests, I still find this solution to be working when applied.

Reference:
http://www.garagegames.com/community/resource/view/11218/


rasteron:
After doing more tests, it seems that other animations like crouch stopped working on multiplayer mode, so I quickly found this solution by @GuyAllard a better approach:

http://www.garagegames.com/community/forums/viewthread/130955/1#comment-831144

PR #1496, opened 2016-01-09
Fix case sensitivity and Platform::fileDelete

camporter:
for linux and OSX. Correct a couple of warnings and errors preventing builds on linux.


SajeOne:
Can confirm this allows compilation of the development branch to complete successfully on Linux. Tested on Arch Linux x86_64.

Lopuska:
Great! Thanks a lot for these linux and mac compile fix!
Yes, we are still not using modern C++ but we should as today it's supported anywhere.
I'll discuss about that with my SC colleagues.
Reviewed and approved as functional, correct and safe PR ;)

PR #1497, opened 2016-01-12
footstep and impact enum extension support

Azaezel:
Removes hardcoded case statements in favor of an offset-driven approach.

While the enums were kept for this for backwards compatibility, I'd strongly advise 4.0 shifts to a more script-driven vector+key approach vs the vector+enum+hardcoded script exposed names.


Azaezel:
The editor doesn't really let you, but fair enough from a manual typo-possibility.

Lopuska:
Cute changes (:

dottools:
Makes a lot more sense this way.

Lopuska:
Works, elegant, code more readable. A+ :P

PR #1504, opened 2016-01-18
ResourceLeakFix for OpenGL

Lopuska:
This resolves some errors related on render targets and other resources which they are never released after creation causing memory leak. It is specific OpenGL fix inside OpenGL render api implementation.
Thank you (:


Lopuska:
This fix is absolutely safe and tested. Changes have effect only on OpenGL render path.

PR #1505, opened 2016-01-18
courtessy @Lopuska: opengl occlusion query fix

Azaezel:


Azaezel:
better version of https://github.com/GarageGames/Torque3D/pull/1441

Lopuska:
Thank you @Azaezel 
Here is a quick description if someone wants to know more about:

This fixes error messages from OpenGL render device during queries requests.
To reproduce this issue, just make sure to use OpenGL rendering, then put a point light in the world with at least 1 opaque object in the scene and set the castShadow flag to on from the point light properties.
The problem is related in the fact that OpenGL has some differences in 'when' we can perform a query to the GPU. In DirectX you can make it at any time, also more than one time in a single drawcall. In OpenGL you can't. In that case, if you have a query active in a target (for example GL_SAMPLES_PASSED), you have to wait that this query finish it's work before you can perform a new one.

If someone want to know how I fixed that, here is a quick but clear explanation:
This is what OpenGL doesn't like: (but it's ok for DirectX)
if ( curEntry.shadowMap )
   curEntry.shadowMap->preLightRender();
if ( curEntry.dynamicShadowMap ) curEntry.dynamicShadowMap->preLightRender(); // argh! can't call it again.

Occlusion query is used here to know if a light is occluded by the geometry. If so, we don't need to render the shadow map.
However we have 2 kind of shadowmap drawed together, 1 static and 1 dynamic.
So instead to store GFXOcclusionQuery object per shadow instance, I move that object inside the ShadowMapParams which is a shared object between static and dynamic shadows. The light has the ownership of this object, so it's perfect to make it our query one single time (not 2) at the beginnning of drawing lights and check it later just to know if we need to render the shadow related to each light.
This has been already tested deeply.

Before the fix, wrong:
http://i.imgur.com/CGpwRsp.png
After the fix, correct result:
http://i.imgur.com/U8oxgEm.png

PR #1507, opened 2016-02-02
Fix for collision issues with scaled players

RoundedIcon:
Players scaled after their creation have collision issues with terrain.
Changing this bit of code fixes those issues for downsized players, even
when shrunk to 10% of their original size.


dottools:
Nice.

RoundedIcon:
misclicked.

PR #1508, opened 2016-02-02
Fix for callOnChildren crashing

RoundedIcon:
callOnChildren crashed my game every time I attempted to use it, so I
tried to fix it. It now works most of the time, but it will still crash the
engine if a method it calls deletes the Simset it is working on.

### DIFF
--- a/Engine/source/console/simSet.cpp
+++ b/Engine/source/console/simSet.cpp
@@ -233,8 +233,13 @@ void SimSet::callOnChildren( const String &method, S32 argc, ConsoleValueRef arg
 {
    // Prep the arguments for the console exec...
    // Make sure and leave args[1] empty.
+   ConsoleValue vals[2];
    ConsoleValueRef args[21];
-   args[0] = method.c_str();
+   vals[0].init();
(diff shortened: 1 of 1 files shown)


dottools:
.safeDelete() method is suppose to be used instead of .delete() to resolve the situation previously causing crashes by attempting to work on a SimObject that was .deleted()'d during the sim tick.

So in a nutshell, only use .safeDelete() and .delete() is only to be used for special circumstances.

If what you're having a problem with isn't related to deleting objects then we need to look into this function you're attempting to fix further and see what's really going on.

RoundedIcon:
To give examples of the torquescript that caused crashes: simset1.callonchildren(delete) < deletes 1-2 things from the set, will crash if you call it a few more times. simset2.callonchildren(deleteSimset2) < calls the method once then crashes. Someone who knows this stuff better then me will have to take a took at those cases.

Lopuska:
Please open an issue for this. and just put a ref on this PR to take trace.
Closing this as not the proper fix.

Lopuska:
fixed on: #1541
You can't delete children object of a delete object.

PR #1509, opened 2016-02-02
Fix fox SFXPlaylist breaking networking

RoundedIcon:
SFXPlaylist tries to write too many bits into the bitstream, which seems
to be responsible for a few issues people have had with non-local
servers. This might not be the best way to go about fixing it, but its
not as bad as what is in the engine right now.

### DIFF
--- a/Engine/source/sfx/sfxPlayList.cpp
+++ b/Engine/source/sfx/sfxPlayList.cpp
@@ -438,24 +438,43 @@ void SFXPlayList::unpackData( BitStream* stream )
    FOR_EACH_SLOT mSlots.mStateMode[ i ]      = ( EStateMode ) stream->readInt( NUM_STATE_MODE_BITS );
       
    FOR_EACH_SLOT if(stream->readFlag()){ stream->read( &mSlots.mFadeTimeIn.mValue[ i ] );}
+		else mSlots.mFadeTimeIn.mValue[ i ] = -1;
    FOR_EACH_SLOT if(stream->readFlag()){ stream->read( &mSlots.mFadeTimeIn.mVariance[ i ][ 0 ] );}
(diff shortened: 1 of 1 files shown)


dottools:
Downside is that you're assuming client side unsent fields are zero'd too which it'll only work reliably initially until the playlist is changed and server tries to re-sync it with the clients again.

RoundedIcon:
Oh yeah, that could be bad. I could stick a bunch of else statements in the read function that set things to their default value. Now that I think about it, I probably should have done that in the first place.

Areloch:
Hey RoundedIcon, need to verify, but have you signed the OSS Agreement? The way it's currently set up at the moment necessitates having an account on the garagegames.com site and signing it there so we can verify it's been signed.

If you have, can you lemme know what your garagegames account is just so we can quickly check that before getting these rolled in? I know it's a pain, but lawyers and liability and all that. We're looking into a few options that'll hook directly into github to make this kinda thing easier, but for now we gotta do it the lame old manual way.

Lemme know so I can start getting your PRs tested and rolled in, thanks! :)

RoundedIcon:
garagegames name is RoundedIcon.

Lopuska:
Thank you for this.
Please, open an issue ticket for this, and track it by pinning this PR #1509 as reference.
I close this as this may not be the proper fix. It's better to find the real cause as your fix could compromise the proper functionality of the SFXPlaylist. Maybe you could fix it just by increasing the bitstream buffer size.
Thank you very much for your contribution @RoundedIcon!

Azaezel:
http://forums.torque3d.org/viewtopic.php?f=10&t=548&hilit=SFXPlaylist#p4651 see forum discusison on the matter.

PR #1510, opened 2016-02-05
Point2/4 constructor Initialize all elements to the same value

rextimmy:
Just fixed up Point2/4 so you can initialize all elements to the same value like in mPoint3I/F/D

### DIFF
--- a/Engine/source/math/mPoint2.h
+++ b/Engine/source/math/mPoint2.h
@@ -203,7 +203,7 @@ class Point2D
    Point2D();                           ///< Create uninitialized point.
    Point2D(const Point2D&);             ///< Copy constructor
    Point2D(F64 _x, F64 _y);             ///< Create point from coordinates.
-   Point2D(F64 xy);                     ///< Initializes all elements to the same value.
+   explicit Point2D(F64 xy);            ///< Initializes all elements to the same value.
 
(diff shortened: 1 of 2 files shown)


rextimmy:
For consistency I really should have added a Point4D class to this.

Engine/source/math/mPoint2.h
```diff
@@ -285,6 +288,13 @@ inline Point2I::Point2I(S32 _x, S32 _y)
 }
 
 
+inline Point2I::Point2I(S32 xy)
+   : x(xy), y(xy)
+{
+   //
```
dottools: Why do these empty comments exist? Why not have it an entirely empty function like what you did in mPoint4.h? I find it cleaner that way.

rextimmy: Yeah cleaner without them but just matching what is already there ;)  https://github.com/rextimmy/Torque3D/blob/mPointX_constructors/Engine/source/math/mPoint2.h#L271-L288

dottools: That's so ugly and pointless. :(

rextimmy: Yeah it's the same story in mPoint3.h too

PR #1512, opened 2016-02-13
[OpenGL - Win32] This fix a bug during resolution change.

Lopuska:
This part I changed was the OpenGL fullscreen support I wrote 1 year ago to help Luis during his efforts.
I formatted the code following the T3D convention, removed legacy headers and I fixed this problem:

If you have Torque3D configured on Windows using OpenGL renderer.
Try this:
-go to main menu.
-set fullscreen checkbox to on. Doesn't matter for resolution you are using.
-set a lower resolution, don't turn back on windowed mode yet.
-return to windowed mode disabling fullscreen checkbox.

You'll return to windowed mode, however the desktop resolution won't be restored as the original.

### DIFF
--- a/Engine/source/windowManager/win32/win32Window.cpp
+++ b/Engine/source/windowManager/win32/win32Window.cpp
@@ -26,8 +26,10 @@
 #include <tchar.h>
 #include <winuser.h>
 #include "math/mMath.h"
+#include "gfx/gfxDevice.h"
 #include "gfx/gfxStructs.h"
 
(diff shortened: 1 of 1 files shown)


Lopuska:
Tested and reviewed. This gives a better support for OpenGL fullscreen on Win32.

Alex-doc:
Everything working on this side.

Lopuska:
@Alex-doc thanks for testing.

PR #1513, opened 2016-02-14
Win32 fullscreen fix partial

RoundedIcon:
A lot of resolutions look like this in fullscreen mode. http://i.imgur.com/DZozand.png
This fixes a few of them.

### DIFF
--- a/Engine/source/windowManager/win32/win32Window.cpp
+++ b/Engine/source/windowManager/win32/win32Window.cpp
@@ -189,8 +189,8 @@ void Win32Window::setVideoMode( const GFXVideoMode &mode )
 		   dv.dmPelsHeight = mode.resolution.y;
 		   dv.dmBitsPerPel = mode.bitDepth;
 		   dv.dmDisplayFrequency = mode.refreshRate;
-			dv.dmDisplayFlags = (CDS_FULLSCREEN);
-			dv.dmFields = (DM_PELSWIDTH | DM_PELSHEIGHT | DM_DISPLAYFLAGS);
+		   dv.dmDisplayFlags = (CDS_FULLSCREEN);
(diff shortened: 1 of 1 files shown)


RoundedIcon:
I did not notice this issue on a different pc when using a nvidia card or intel integrated graphics. This might be an issue with amd graphics cards only.

RoundedIcon:
More testing - turning on gpu scaling in Radeon settings fixes / hides the issue. This pull request hides the issue for some resolutions when that is not on.

dottools:
I see no problem with these changes provided no negative side effects.

RoundedIcon:
To clear something up - The existence of the black bars is not the issue. The problem is that they appear on resolutions with aspect ratios that match the monitor. This isn't meant to be a final fix, it just makes the deeper engine problem of inconsistent bar behavior slightly less problematic.

Lopuska:
In which render backends is present this issue? Because the query/enumeration of display modes differs between ogl and d3d.
See GFXGLDevice::enumerateVideoModes() and GFXPCD3D9Device::enumerateVideoModes()

And also new resolution change is handled differently.
in opengl we use ChangeDisplaySettings api and in directx a direct call to swapchain:
swapchain->SetFullscreenState()

I'll take a look in deeper way soon! (:

RoundedIcon:
The problem happens with both d3d and ogl. Ogl has an additional problem as well - Increasing the resolution in full screen mode only changes the screen res. The game window starts shrinking into the upper left corner with large res increases.

Lopuska:
I'm taking a look about this now. Thanks. Also this fix a different problem relative to fullscreen:
#1512

Lopuska:
I close this as this is not fixing the issue. Also the problem should be fixed in a different way.
Please, open an issue ticket for taking trace of this! I'll try to investigate it after other priorities.
Thank you for your contribution. (:

Issue #1515, opened 2016-02-15
‘vsprintf’ was not declared in this scope


practicing01:
Merged latest development: /Engine/source/core/stream/stream.cpp:131:52: error: ‘vsprintf’ was not declared in this scope
    const S32 length = vsprintf(buffer, format, args);

Azaezel:
try const S32 length = dVsprintf(buffer, sizeof(buffer), format, args);

practicing01:
Ok, after that I get /Engine/source/math/mathUtils.cpp:1850:90: error: explicit qualification in declaration of ‘void MathUtils::mBuildHull2D(Vector<Point2F>, Vector<Point2F>&)’
 void MathUtils::mBuildHull2D(const Vector<Point2F> _inPoints, Vector<Point2F> &hullPoints)

practicing01:
Ok fixed that by removing "MathUtils::" from the definition.

PR #1516, opened 2016-02-16
vsprintf replacement with engine vairant

Azaezel:
resolves first issue in
https://github.com/GarageGames/Torque3D/issues/1515#issuecomment-184446719


Lopuska:
Correct! thanks Az.

PR #1520, opened 2016-02-17
add a "no weapons" third person cam mode for player

blackwc:
This adds a new camera mode in third person for the player. It's meant for when the player has no weapons (or the weapons are holstered). The camera sticks to the back of the players when moving otherwise it can freely rotate around the player.

Thanks to "newaged" from the IRC for figuring out a rotation issue.

As of right now this is only for the standard move. I plan on adding support for extended move.

It works in script similar to how freelook works. If $mvNoWeaponsTPCam is true then the no weapons camera mode is used for third person. This can turned off at any time by setting $mvNoWeaponsTPCam to false.


Engine/source/T3D/player.cpp
```diff
@@ -2641,10 +2641,33 @@ void Player::updateMove(const Move* move)
          }
          else
          {
+
             mRot.z += y;
-            // Rotate the head back to the front, center horizontal
-            // as well if we're controlling another object.
-            mHead.z *= 0.5f;
+
+            if(move->noWeaponTPCam) {
```
dottools: starting braces should be on their own line and vertically aligned with ending braces just like the rest of the source file.

rextimmy: I stopped counting at 40 other occurrences in this file of incorrect braces, hardly fair to pick on him.

blackwc:
@rextimmy thanks though I prefer your unedited comment where you said it's hardly fair to pick on me! :P 

Anyway, I will use proper braces when I add this to extended move. However, I'm certainly not going to fix every single case of incorret braces.

Engine/source/T3D/player.cpp
blackwc: See comment [here](https://github.com/GarageGames/Torque3D/pull/1520#issuecomment-185796347)

RoundedIcon:
Now is probably a good time to mention that this will break networking. In player.cpp in the packupdate function, you will find this line.
`      stream->writeSignedFloat(mHead.z / mDataBlock->maxFreelookAngle, 6);`
writeSignedFloat does not like values over 1, which these code changes can give it. It could be changed to use the the largest value between pi (2pi?) or maxFreelookAngle. Also, brackets :p.

blackwc:
I welcome review and testing but hold off on any merging because I need to add more stuff to make it more like the "standard" third person cam controls.

Areloch:
Alright, sorry for the delay on testing this.

Tried it out. Looks like the $mvAltFreeLook works(though I'd recommend having an actual keybind to make it occur rather than needing the console action, just for example usage cases if nothing else.

The main problem I noted was that when you kick it on and start looking around, it's nice and smooth on the client side for the client doing the looking, but it doesn't seem to be getting transited properly to other clients, as the free looking player's character begins to twitch back and forth, at range relative to how far around they've orbited the camera.

If that can get fixed, this seems to work pretty well otherwise

PR #1525, opened 2016-02-20
Added immutable vertex and index buffers.

TheDushan:


Lopuska:
Testing it now! ;) Thanks for this, really useful for lockless rendering cycles!

Lopuska:
Good improvement, it'll help also in the perspective of the GFX2.

PR #565, opened 2014-01-24
Seamless terrain.

signmotion:
Heights, sizes and materials on the edges of terrains are corrected.

Before patch (4 terrains)
image

After patch
image

[Test level](https://drive.google.com/file/d/0B4G7HDT9STV-ZDFrUmg3eDd5eHc/edit?usp=sharing) (extract to the folder 'game/levels/test'). Smart run from the command line
> application.exe -level "test/seamless-terrain/test1.mis"

PR #954, opened 2014-11-20
Complete rewrite of  https://github.com/GarageGames/Torque3D/pull/566

Winterleaf:
Removed all the C++ 11 code
Fixed a bug where you couldn't paint the last pixel on the edges.

PR #1531, opened 2016-02-26
Seamless terrain. #565

Lopuska:
Some correction of #565


Lopuska:
I close this as PR #954 contains also these additions!

PR #1536, opened 2016-02-29
opengl crashfix pow(x,y) needed to be passed matching vartypes.

Azaezel:


Lopuska:
Thanks Az! what's the reproduction case?

Azaezel:
Showed up in the latest ubuntu review of devhead.

Lopuska:
thanks! (:

Issue #1537, opened 2016-03-01
specular sliders are switched


Alex-doc:
I've just noticed that the specular sliders are now switched:
[shadergen](https://github.com/GarageGames/Torque3D/blob/908be4818f7e5de23f2070d18ed4df0f4dfc5ddf/Templates/Full/game/tools/materialEditor/gui/guiMaterialPropertiesWindow.ed.gui#L2004)
[editor slider value](https://github.com/GarageGames/Torque3D/blob/908be4818f7e5de23f2070d18ed4df0f4dfc5ddf/Engine/source/lighting/advanced/hlsl/deferredShadingFeaturesHLSL.cpp#L163)

Same for the specPower should be /128 in shaderGen and vice versa

Lopuska:
Thanks Alex ;)

Lopuska:
Fixed by @Azaezel

PR #1539, opened 2016-03-02
addresses https://github.com/GarageGames/Torque3D/issues/1537 via the…

Azaezel:
… following:

https://github.com/GarageGames/Torque3D/blob/908be4818f7e5de23f2070d18ed4df0f4dfc5ddf/Engine/source/materials/materialDefinition.cpp#L261 denotes power as sharpness, reflected in the size/falloff of a highlight halo *

https://github.com/GarageGames/Torque3D/blob/908be4818f7e5de23f2070d18ed4df0f4dfc5ddf/Engine/source/materials/materialDefinition.cpp#L264 denotes strength as overall brightness of highlights. **

*and sharpness of reflection if using a cubemap.
**  reflected in cubemapped objects as also degree of 'reflection' vs diffuse/albedo coloration.


Lopuska:
Thanks Az! I told alex to make an issue for this yesterday as we were together on skype call ;)

PR #1540, opened 2016-03-04
Fix zipped dts loading

Alex-doc:
fixes the loading of cached dts shapes on zip(#1466 ):
if the cached shape is not present, it warns through console and forces it to be cached to disk.
If the dts is present in the archive, it will be used (unless forceDAE) regardless of the creation date, since it was zipped on purpose.


blackwc:
Seems fine.

rasteron:
Hey guys, this works great! I got to try it on static models with just a material cs file. Have you tried it with scripted objects like packing a cached dts weapon folder? I'm getting `animation sequence not found` errors and it searches for the DAE file version again when packed.

PR #1541, opened 2016-03-04
Fix crash callonchildren

Alex-doc:
Anis and me managed to fix this in a proper way than: #1508


Lopuska:
This just prevent a crash caused by callOnChildren call.
You shouldn't call delete function directly. You can do it with a schedule:
exampleGroup.callOnChildren(schedule,0,"delete");

PR #1542, opened 2016-03-04
[solder_terrain_edge] rewrite of #954

Lopuska:
Adds OpenGL support, fixed tabs vs space, avoid standard c++ container, using T3D native types.


dottools:
What's up with all the file renames? Think you're going to have to take your PR as a patch on top of a cleanly cloned `Development` branch and apply only the source code changes with said patch. Cause something went weird with commit tracking on these changes.

Lopuska:
@dottools Sorry for confusion, wasn't my intention to open PR in that early stage (I unintentionally created it from the git client). I still was working on it...
However, now it should be ok.
It just needs a final review.
Thank you (:

PR #1545, opened 2016-03-08
Correcting an outlier case with animated materials breaking.

Areloch:


rextimmy:
I have a fix for this already in the dx11_deferred if you want to wait for that jeff? Up to you though.

Areloch:
Yeah, DX11 is going in shortly, so we'll just wait with that there. I'll close this.

PR #1547, opened 2016-03-09
CMake torqueConfig.h.in fix

rextimmy:
A few parts of the engine check for _TORQUECONFIG_H_


Areloch:
Looks like this needs another #endif tacked onto the end to make it work, otherwise we get a mismatched pair.

Also, do we need to remove the pragma once? I remember seeing something somewhere(may have been Hal in the irc) talk about how having both the ifdef and the pragma can help cover more cases to improve compile times.

rextimmy:
Oh shit i left off the #endif , oops.

PR #1550, opened 2016-03-15
new method: ResetGFX();

Azaezel:
exposes the GFX->beginReset(); method to script to allow folks to force the gbuffer to reinitialize (if, say a custom element is holding on to data in a buffer and it needs a cleaning, to name one example)


dottools:
No doc string? No newline at end of file? What year is it?! ;)

PR #1558, opened 2016-03-19
Spacial update for Px3 character controller

OTHGMars:
The testSpacials() and setSpacials() functions have been added to test for overlaps before changing character poses and to update the character controller with the new dimensions.


Lopuska:
For me it looks good and correct.
Thanks for contribution.

PR #1559, opened 2016-03-20
Direct3D11 Support

rextimmy:
Contributions from the following:
* Anis
* ZeroFault
* Azaezel
* Andrew Mac

Sorry if I left anyone else off that list. 
Time for performance phase 2 now.


Lopuska:
yeaaaaaah!!! finally @-@

dottools:
Awesome. Looking forward to trying it out.

JeffProgrammer:
grats guys!

Areloch:
Initial testing last night checked out. Will give it another run through tonight before giving it my seal of approval for merging.

jamesu:
There is a big problem with this code with regard to the implementation of the clear method. 
This is what is used to clear the view:
	if (flags & GFXClearTarget && rtView)
		mD3DDeviceContext->ClearRenderTargetView(rtView, clearColor);

Unfortunately the D3D11 docs state the following:

"Differences between Direct3D 9 and Direct3D 11/10: Unlike Direct3D 9, the full extent of the resource view is always cleared. Viewport and scissor settings are not applied."

Ergo when the clear method is used when a viewport is set, the whole render target is cleared which is causing problems for stereo rendering, and quite possibly anywhere else which relies on clear to not clear the whole view (like some parts of the gui).

rextimmy:
Yes james is correct, a fix is on it's way.

Issue #1561, opened 2016-03-23
Linux Binary Dist includes only Windows Executables


italic-r:
I downloaded the 3.8 Linux binary distribution and extracted it only to find Windows executables. Is this intended? I wanted to build it locally, but it requires cmake 2.8.12, and Centos 7 is locked to 2.8.11. Getting cmake built and integrated with my system is far more trouble than it's worth and potentially breaks the system, so I need to avoid that.

Thanks

TheDushan:
Easy way is to manualy change hardcored version inside "CMakeLists.txt"

Inside "Torque3D/CMakeLists.txt" you have there "cmake_minimum_required (VERSION 2.8.12)
". That is on line "1", you cannot miss it.
lower that to version what your cMake for CentOS is supporting and that will solver your problem.

Thanks for reporting this, I hope that someone with write access will edit "CMakeLists" and fix that
:)

Italic-:
The issue with that is some build systems require features introduced in 2.8.12. I've tried to build other projects, like Open Broadcast Software, by lowering the required cmake version, only to run into unavailable or unsupported features. I can't seem to find a full version of the changelog for 2.8.12, but this is as close as I can get: https://cmake.org/pipermail/cmake/2013-October/056020.html

I'm only an amateur builder and don't know the specifics of cmake. Someone more knowledgeable of Torque's build system would have to tell me if 2.8.12-specific features are being used.

TheDushan:
No.
AFAIK, none of improvements what are introducted with "2.8.12" versions are used while generating build solution
Also, others used "2.8.12" version because they thought that it is the lowest known working version number for all platforms.

JeffProgrammer:
If i were you I would go download a newer CMake from CMake.org. Obviously this is the problem with Linux package managers. They do not keep up to date on stuff. We are now at Cmake 3.5 yet for some reason the package manager you are using is still on 2.x. I would personally advocate for completely dropping 2.x support as it is old.

Areloch:
Yeah, I was just about to link https://cmake.org/download/

I'd say go and try and grab and use the latest binary from there first before having to worry about hacking in support for older versions. 

If that doesn't work, we can take it from there.

Italic-:
I didn't realize I could download premade binaries for cmake... 

[...]

PR #1566, opened 2016-03-24
Fix release build compile with MSVC 2015 (finally)

JeffProgrammer:
This fixes compiling a release build with MSVC 2015. The issue was that the DirectX runtime requires a function that is no longer present in MSVC 2015. However, Microsoft was nice enough to add a legacy compatibility library to continue using it, so linking against that fixes the linker errors and it works again.

However, we have to check for MSVC 2015, and to do that requires CMake 3.1.4 as that is when they added MSVC14 macro. If that is not okay, I can come up with a solution to keep backwards compatibility, but c'mon its 2016. They are at CMake 3.5.0 now :/

### DIFF
--- a/Tools/CMake/CMakeLists.txt
+++ b/Tools/CMake/CMakeLists.txt
@@ -20,6 +20,9 @@
 # IN THE SOFTWARE.
 # -----------------------------------------------------------------------------
 
+# JTH: We require CMake 3.1.4 for MSVC14 compatibility check
+cmake_minimum_required(VERSION 3.1.4)
+
 include(basics.cmake)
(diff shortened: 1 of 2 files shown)


dottools:
JeffH is still alive! And thanks for the PR. ;)

JeffProgrammer:
Had some free time today :)

Areloch:
Works great.

I'll wait on getting a reply from Italics about the cmake binaries thing for issue #1561 just make as informed decision on this as possible, but this will likely get rolled on in shortly.

Thanks for the contribution, man :)

PR #1573, opened 2016-04-04
allows navmeshes to generate for most scene objects, 

Azaezel:
and adds a NavMeshIgnore method for object-instances to filter them out.

(primary usage this end: bots pathing around one another via a cyclic call to the previsouly turned over NavMeshUpdateAll/NavMeshUpdateAroundObject)


eightyeight:
FWIW, the long-term solution here is probably to implement more of Detour's path following and crowd avoidance features, which I believe includes dynamic avoidance.

I don't like adding pathfinding specific info to SceneObject, but it's a pragmatic solution that's probably the best possible result without going the full Detour route, which will take significant time/effort.

PR #1574, opened 2016-04-06
corrects projection matricies for opengl

Azaezel:


Duion:
Tested it and it fixed the issue for me.

PR #1576, opened 2016-04-07
Updates SDL to version 2.0.4, which makes it compatible with VS2015.

Areloch:


dottools:
Something really wrong with this PR.

dottools:
Nevermind, it's just removing a bunch of junk/unnecessary files we shouldn't be distributing anyway.

Areloch:
Heh, yeah, this PR also removes out stuff like the android and ios stuff, because as-is, it's totally useless to us.

May go back in later, but for now it's just bloat. ;)

JeffProgrammer:
GL issues:

[12:38:44 PM] Jeff Hutchinson: on outpost constant console spam of opengl issues with GFXFormatR16F and GFXFromatR16G16B16A16F
[12:38:53 PM] Jeff Hutchinson: says "not supported with specified profile"
[12:38:57 PM] Jeff Hutchinson: which probably means they aren't part of core

ALSO

FXAA is cropping up, won't compile the shader says it needs SM 5.0 arb extension or #version 400

Azaezel:
Can confirm https://github.com/Areloch/Torque3D/tree/GLCoreExtensionFormatFix corrects the above format lookup glitches

JeffProgrammer:
clarification on my comment: the GL issues in my comment are on the current dev branch too. so the branch az mentioned needs to be merged in and then that fixes it.

PR #1580, opened 2016-04-12
Initial commit for the new base template.

Areloch:
Replaces the Empty and Full templates with a new, re-factored Base Template.


dottools:
Are files actually being moved from FullTemplate to BaseGame or is github confused on this PR?

Areloch:
Empty and Full are being replaced in favor of this.

I wanted to make sure the PR got made before my lunch break concluded, I'll have a writeup on what all this actually impacts here in a bit.

dottools:
Wouldn't it be better as an addition to instead of replacement? Considering the latter is more disruptive than the former, especially to those who already have existing T3D projects/games in development.

Areloch:
Except that leads to even more duplication than we already have. Point of this was to revamp the core, streamline it out and cut out all the unneeded duplication of everything.

I'm drafting a wiki page on how to port over projects to work with this.

Areloch:
Did some discussion with Az and rextimmy and a few others and we all seem to agree that given the big shift this has, it's better suited for 4.0, where some breakage of backwards compatibility will be more acceptable.

That, and there's some configuration/layout changes that need to be done on this to make it better, so pushing it back is doubly practical.

dottools:
I know that `development` branch is typically used for actual development changes to the engine, but would it be more useful to have say a `dev4.0` branch to go ahead and put this PR into, in order to make improvements along the way reaching towards v4.0? While `development` branch can stay on track for v3.9?

eightyeight:
I tried pushing for more dev branches, but it was very confusing for lots of people. If GitHub let us retarget PRs to new branches there'd be no problem.

Also, just to put it out there,  4.0 doesn't have to come after 3.9. There could be a 3.10, 3.11 etc. if necessary. I was just a little suspicious how 4.0 was created immediately after 3.9... these are versions not decimals ;)

John3:
I think is good idea to have other branch, and separate things and start pushing to v4... dottools have a point

I was one of the confused people went I start using torque3d... but was because was not documented anywhere the meaning of the branches.

This is just an idea... other approach we can use, is keep master and development, do the merge as usual and release as usual, went we start merging v4-commits stuff, to keep stuff clean, made another branch dev3.9 to separate thing. At this point development and dev3.9 have the same thing. Now start merging of v4-commits stuff to the development branch.

Why a branch for 3.9? because with v4 we know that some stuff don't have backward compatibility, want or not, I think is good idea to have another branch at some point. This approach I think will be helpful to old projects.

PR #1591, opened 2016-04-26
Re-enable destruction of PhysicsShapes.

Areloch:
Adds in some missing script functions that let projectiles damage and destroy physics shapes.


TylerD7:
Adding these to functions to my project allowed objects to be destructible!

PR #1594, opened 2016-05-06
Fixes the cmake blacklist filter it not using SDL

Areloch:
Fixes the blacklist filter in the event we aren't using SDL to avoid including an unwanted file dialog codefile.


JeffProgrammer:
can confirm working now here in MSVC 2015

PR #1595, opened 2016-05-06
Rolls back OGL Projection correction.

Areloch:
Epoxy looks to handle the projection depth range so it behaves more like D3D, so this change was doubling up and causing problems.


JeffProgrammer:
I can confirm that this fixes the OGL shadows.

PR #1597, opened 2016-05-07
OpenGL Extension Caching

JeffProgrammer:
This caches OpenGL extensions that are not part of the core profile. It uses a struct of Boolean values to avoid function call overhead as @rextimmy suggestion. Extensions that run during initalization are not cached, as there really is no reason to as of now. It is fairly easy to add more extensions to the cache as needed. 

This PR also removes the old extensions that are no longer needed.

Finally, it replaces a call of glGenMipmapEXT to glGenMipmap, as we are in the core profile now!

### DIFF
--- a/Engine/source/gfx/gl/gfxGLCardProfiler.cpp
+++ b/Engine/source/gfx/gl/gfxGLCardProfiler.cpp
@@ -65,23 +65,23 @@ void GFXGLCardProfiler::setupCardCapabilities()
    // Check for anisotropic filtering support.
    setCapability("GL_EXT_texture_filter_anisotropic", gglHasExtension(EXT_texture_filter_anisotropic));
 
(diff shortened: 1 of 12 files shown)


JeffProgrammer:
Also I meant 3.2 on my description of the last commit, sorry!!

dottools:
Uh oh tabs vs spaces indention strikes again. ;)

Areloch:
Seems to work fine.

JeffProgrammer:
great to hear!

BadFactor:
Wich Version of OpenGL is it ? If I'm not wrong, there isnt any description of it!

JeffProgrammer:
@BadFactor OpenGL 3.2 core profile is the min. required to run Torque3d.

PR #1599, opened 2016-05-07
Added missing OpenGL profiling blocks.

JeffProgrammer:
Please check to make sure I put these in the right places. There's a couple that I had to leave out from the D3D9 ones because they didn't make sense, but I tried to match as many as I could.


BadFactor:
Thumbs up!

PR #1602, opened 2016-05-08
Correct SpinLight animation. https://github.com/GarageGames/Torque3D/…

John3:
…issues/1601


Areloch:
Az's change with  #1604 allows a little more flexibility in controlling the animations if we add more options going forward, so I went with that, but for the record, this fix did indeed work as well, so thanks John :)

Issue #1486, opened 2015-12-18
Crash When Selecting Camera And Material Editor Is Open


SteveYorkshire:
Does exactly as it says on the tin.
When in the editor in player view and you try and select the camera, Torque3D will crash if the Material Editor is open. Not really an important bug, just an annoying one when you forget to switch out of Material Editor first.

John3:
Hello @SteveYorkshire 

I tested this bug in a new clean t3d development branch, and I don't have this issue. Please can you see the video and tell me if correct the steps I do?

<a href="https://drive.google.com/file/d/0B9GUwtDHCp_eQV9kZmVhbFM0Tlk/view?usp=drivesdk">https://drive.google.com/file/d/0B9GUwtDHCp_eQV9kZmVhbFM0Tlk/view?usp=drivesdk</a>

If is correct can we close this? @Areloch

Areloch:
Took a look at this m'self, and i got the crash to occur and PR'd a fix. If you or steve can test and affirm it, I can roll it in.

John3:
Ok, I modified Engine/source/T3D/shapeBase.cpp as you can see in the following image. Clean environment, new project, last development branch....

<img src="http://i.imgur.com/T96PUdG.png"/>

And as you can see in the following image ended correctly.

<img src="http://i.imgur.com/54lk8YB.png"/>

I did the same test without issue.

Areloch:
Fixed per the PR.

PR #1607, opened 2016-05-09
Select camera when in material editor crashfix.

Areloch:
When attempting to select a camera Object(so going into the editor, going into the freefloating camera and then back to the player camera and selecting the freefloating one) while the material editor was open, it would crash because it attempts to access the shapebase object's shapeInstance - but cameras don't have a shape.

This fixes it so it makes sure there's a shapeinstance before trying to get the material data.

Pertains to #1486


dottools:
This is awkward.. spaces vs tabs indention conflict. ;)

Areloch:
I corrected the code I changed :P

SteveYorkshire:
Looks good, no crash. :)
I get a parse error in console but I reckon that's because camera has no material.

Areloch:
Yeah, that'd be a separate thing to fix. Was mostly concerned with the crash ;)

PR #1612, opened 2016-05-15
Confirming argv[0] has a non-zero value

shrubchicken:
This little if statement just confirms that argv[0] got a non-zero value. Since as is a junk value can be concatenated with .so and dlopen will search for the file, this helps prevents that. It shouldn't slow down execution most the time too much, and should help in cases where something has gone wrong. The printf value could be changed to something others find more appropriate.

### DIFF
--- a/Engine/source/main/main.cpp
+++ b/Engine/source/main/main.cpp
@@ -240,6 +240,8 @@ int main(int argc, const char **argv)
    
    if(strlen(argv[0]) != 0) {
    int len = strlen(argv[0]);
+   } else {
+      return -1;
    }
    
    char *libName = new char[len+4]; // len + .so + NUL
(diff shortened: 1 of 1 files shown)


shrubchicken:
Sorry, this version doesn't have a printf statement. I can fix that if needed. I wasn't sure what would be appropriate here anyway. My apologies.

dottools:
This would have been better to insert as it's more to the point:
```C++
// abort on no bin name provided
if(strlen(argv[0]) == 0)
   return -1;
```

shrubchicken:
So we're getting argv[0], which possibly has no value and relying on that strlen? This uses less code, sorta (you're still getting the value from argv[0] either way), but it's not really more to the point?

PR #1614, opened 2016-05-16
Update coverPoint.cpp

shrubchicken:
Assuming you can't peek in more than one direction at a time, in cases where the peek direction isn't "over", this'll decrease the amount of checks performed. There were previously three if statements, the program would have to run through the other ones even if the first one was satisfied.


dottools:
That is not how peeking works as it isn't a single state, but three individually configurable states.

You're modifying the rendering of cover points for recast path finding which only draws while in world editor using the navigation editor, where it's important to have a visual aid of what the cover point's configuration is anyway.

shrubchicken:
Alright, sorry about that. Thanks for the heads up.

shrubchicken:
I'll go ahead and close the pull request. Couldn't decide if the branch should be deleted, but I'll go ahead and delete it too. I would have closed it sooner, didn't realize I could. Thanks for the clarification though, glad I didn't break anything.

PR #1615, opened 2016-05-18
Entity/Component implementation

Areloch:
Adds a (experimental-flagged) Entity/Component system to Torque 3D.

Can be activated via the advanced property TORQUE_EXPERIMENTAL_EC in cmake.

Note: Currently requires one of the Physics APIs(Bullet or PhysX) to be active as well to use any of the physics/collisions components.


Areloch:
Nah, you're correct. Timmy recently pointed that out to me and it had totally snuck past. My apologies and I'm working on the fix for it. Thanks for the additional spotter on that though!

PR #1617, opened 2016-05-19
removes w=z trick (was causing fisheye, effectively)

Azaezel:


jamesu:
BTW I can confirm this fixes the vomit-inducing warping effect that can be seen through a VR HMD.

PR #1620, opened 2016-05-20
Bug space folder in scene tree 2

John3:
Bug space folder in scene tree

Fix by David Robert Pemberton
https://www.garagegames.com/community/blogs/view/22295

Before, the space break up the folder in to 2 folders.

Bug fix: Now you can see the folder "soldier actor" as one folder.

### DIFF
--- a/Templates/Empty/game/tools/shapeEditor/scripts/shapeEditor.ed.cs
+++ b/Templates/Empty/game/tools/shapeEditor/scripts/shapeEditor.ed.cs
@@ -380,7 +380,8 @@ function ShapeEdSelectWindow::navigate( %this, %address )
 
       // Ignore assets in the tools folder
       %fullPath = makeRelativePath( %fullPath, getMainDotCSDir() );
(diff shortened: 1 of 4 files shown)


Areloch:
Nice :D

Two things: For the future, you can just do another commit to the same repository with a correction if it's something like fixing a tabs issue or whatnot. You don't always need to scrap the PR so much as just update it :)

Second, figure I'd have people weigh in on this as it's an opinions thing, but would it be better with an underscore? Or leave the | pipe in there? I get the feeling an underscore may be easier to understand?

Like said, open for opinions on this. If we're fine with the pipe, I can merge it in as-is.

John3:
thanks for the tips @Areloch I send other PR 

before


after
<a href="http://i.imgur.com/fQuwy36.png">http://i.imgur.com/fQuwy36.png</a>

John3:
:O sorry! my internet played with me... and for some reason is closed!

Issue #1630, opened 2016-06-06
Bob.setImageTrigger(0,1); does not rapid firing


John3:
This command tells the weapon to start rapid firing, but it's only shoot once.

`Bob.setImageTrigger(0,1);`

Last working commit:
0bf021280e75d5d213da4363874958ff2226b882

Caused by:
02c88201c67601caedea320a42e96643e8361a02

Areloch:
Looking at the code from the change that caused the problem, a better approach to this may be to have the setImageTrigger console command overriden in AIPlayer, which sets the relevent move trigger that way. 

Looks like the intent of the original change was to be able to set regular move triggers directly, such as the jump move trigger, etc, without needing to go through a mounted image(in the even the AI doesn't have one).

If we override the setImageTrigger which would then call setMoveTrigger, that'd let us keep the ability to activate AI move triggers without a mounted image, while still letting said mounted images activate triggers if required.

Thoughts?

John3:
> the setImageTrigger console command overriden in AIPlayer

Hey hi @Areloch  Where exactly? out of "bool AIPlayer::getAIMove(Move *movePtr)" https://github.com/GarageGames/Torque3D/blob/ca4e4fe86d57f15e8bf7f8eb25fd58ab1f327231/Engine/source/T3D/aiPlayer.cpp#L348

BTW just commenting here.... the intended functionality of "setImageTrigger(0,1)"  to my knowledge is the rapid firing, and was documented this way. 

Before was working the rapid firing, after the change only shot once, so I reported as a bug. This is a bug? or this is the new functionality? ty

PR #1631, opened 2016-06-06
Fix for rapid firing setImageTrigger(0,1); issue #1630

John3:


John3:
yea you right I was doing very early in the morning before go to work and was sleepy! xD

Issue #1633, opened 2016-06-06
Crash on "add a new mesh" Forest Editor action


dakk:
In the world editor -> forest editor, when I try to "add a new mesh", the software crashes in this assertion: https://github.com/GarageGames/Torque3D/blob/development/Engine/lib/nativeFileDialogs/nfd_gtk.c#L51

John3:
Hi @dakk , how I can replicate the crash? I tested in a clean environment and I don't have this issue. I added two kind of tree and them go to the brushes tab and use it

<img src="http://i.imgur.com/O5dIFdn.png"/>

dakk:
The crash happens only on Linux, the assertion is on the gtk nativefiledialog; I just tried with the empty template and it crashed. The steps to reproduce are:
1. Open the world editor
2. Click on Forest Edtior
3. Switch to mesh tab
4. Click on the "add mesh dialog" 

And then:

``
% TestEmpty: /home/dakk/Repositories/Torque/Torque3D/Engine/lib/nativeFileDialogs/nfd_gtk.c:51: AddFiltersToDialog: Assertion `strlen(typebuf) > 0' failed.
Aborted
``

PR #1634, opened 2016-06-06
Fix for "Crash on 'add a new mesh' Forest Editor action" #1633

dakk:


Areloch:
Seems to work. Wonder where the issue was coming from though, seemed to work fine with other file types.

Areloch:
As this works, I'm going to go ahead and merge it, but keep the issue open as a general reminder to look into the odd behavior. May just be an oddity with the lin side of the NFD as windows doesn't see the problem. Worth keeping a note on at least.

PR #1640, opened 2016-06-09
fallback - dynamic shadow cache nuking (static refresh delay preserved)

Azaezel:
worst case scenario fallback for if we can't track down why vector lighting seems determined to shift positions periodically based upon some influence by the dynamicrefreshfrequency rate


Azaezel:
Will try to find a smarter solution, but worst case scenario this'll at least not mess with folks transitioning  versions.

PR #1642, opened 2016-06-09
level-wide accumulation assignment 

Azaezel:
* overidden if an object's origin is in an accumulationVolume.


Areloch:
Tested, and we need to add a 

`
if (!mAccuTexture.isNull())
   {
      mAccuTexture.free();
      gLevelAccuMap.free();
   }
`

into LevelInfo's deconstructor. Can get texture leaks on close if we don't.

Also, I had to nudge a shape to get it to apply, even on a reload. It sounds like you only had to poke it when editing and a it worked fine on reload, right?

PR #1644, opened 2016-06-10
Fixed compile errors on linux (obvious programming faults).

Bradan:
A bool is not a pointer and a reference has to be returned.


eightyeight:
Thanks for that, @Bradan! I'll leave it to the regular contributors to evaluate but it looks good to me :).

PR #1654, opened 2016-06-17
Shape mounting

Azaezel:
work by @OTHGMars


Azaezel:
with apologies for being a slacky McSlackerson there.

PR #1658, opened 2016-06-21
Corrected "Pref::Server::ConnectionError" description.

John3:
Change "the FPS starter kit" for "Torque3D".


Azaezel:
suggestion:
https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/game/main.cs.in#L24

PR #1661, opened 2016-06-22
Enable demo rec

John3:
This is how looks the new menu. Now is possible to play the recorded gameplay.


Areloch:
This looks good. I'd just want the delete button working before this gets rolled in. Get that, and I think it's good.

PR #1678, opened 2016-07-09
Fix a few more case sensitivity issues

Areloch:
Fix a few more case sensitivity issues for Linux.


John3:
I don't see the changes for the empty template

John3:
you need a hand with the empty template?

PR #1681, opened 2016-07-10
Added a test physicsShape datablock

John3:
Added a physicsShapeData datablock to use as template and shape test.

This will enable an easy way to add a physicsShape from datablock editor. 

If this test object don't exist, you can't add a new PhysicsShapeData, the new datablock need an object to copy values from. Unless you made the script by hand and then load to your level, but again you need to copies values from somewhere.. new peoples not known this values. This wil help to take down this boundaries.

1) Before you need to compile with bullet or physX
2) Uncomment from art/datablocks/datablockExec.cs
`//exec("./physicsShape.cs");`

Now you can add with a few clicks a physcisShape

1) Create

2) Edit

3) Add

### DIFF
--- a/Templates/Full/game/art/datablocks/physicsShape.cs
+++ b/Templates/Full/game/art/datablocks/physicsShape.cs
@@ -20,48 +20,22 @@
 // IN THE SOFTWARE.
 //-----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


John3:
closed this, the fix need to be done in other way. #1689

PR #1684, opened 2016-07-11
update Bullet from 2.82 to 2.83.7

John3:
this PR already have this #1683

I tested and it seems to work without issue.

 Can anyone double check? :D


Areloch:
Woah, nice work. I'll definitely be checking this over.

rextimmy:
Nice work, tested ok here.

John3:
**PR cleanup_1:** remove bullet cmake and premake files.

Lopuska:
It's cool, but I have a finished and already tested Bullet 3 port. maybe we can considering to switch directly to the last version instead of 2.83?

John3:
That's greats lopuska, BTW when you will do a PR with bullet v3? will be in months? Because this v2.83 can included right now without any problems.... even you could be added before the release of v3.9

Lopuska:
@John3 If for you it's a big priority, I can manage Bullet 3 PR as soon as possible. Also I'll need a reviewer to check my code before proceed with merge. Do you want to be the reviewer? :)

@Areloch Do you think it's good to switch directly to Bullet3?

John3:
well my main complaint if you think add bulletv3 right now go ahead of course anis :D if not... why you can't add a bullet upgrade? the actual version 2.82 is from 2013, the v2.83.7 is from this year 2016.

John3:
hey hi! :) this still considered?

PR #1688, opened 2016-07-12
Basic OpenVR Support code

jamesu:
Adds basic OpenVR support. This also includes support for rendering controllers based on OpenVR data, though this is as of yet incomplete as there are no hooks for the button input.

Much like the oculus support, a script is also provided to handle setting up control input. However this is not hooked in by default for this PR.

The intention of this PR is more to add the support code to the engine and as a basis for identifying regressions with future dx11/gl changes.

### DIFF
--- a/Engine/source/T3D/player.h
+++ b/Engine/source/T3D/player.h
@@ -686,7 +686,6 @@ public:
    void getEyeBaseTransform(MatrixF* mat, bool includeBank);
    void getRenderEyeTransform(MatrixF* mat);
    void getRenderEyeBaseTransform(MatrixF* mat, bool includeBank);
-   virtual DisplayPose calcCameraDeltaPose(GameConnection *con, const DisplayPose& inPose);
    void getCameraParameters(F32 *min, F32 *max, Point3F *offset, MatrixF *rot);
    void getMuzzleTransform(U32 imageSlot,MatrixF* mat);
    void getRenderMuzzleTransform(U32 imageSlot,MatrixF* mat);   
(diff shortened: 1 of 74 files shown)


Engine/source/T3D/player.h
```diff
@@ -39,6 +39,7 @@ class DecalData;
 class SplashData;
 class PhysicsPlayer;
 class Player;
+class OpenVRTrackedObject;
```
Azaezel: Causes non-compilation with a default-selection build. Initial suggested resolution:
#ifdef TORQUE_OPENVR
class OpenVRTrackedObject;
#else
class OpenVRTrackedObject : public GameBase
{
   typedef GameBase Parent;
};
#endif

jamesu:
@Azaezel that sounds pretty bad since your giving the fake impression that OpenVRTrackedObject is actually implemented according to the dummy stuff in the header, which could cause problems depending on the order of includes.

Could you at least tell me what compiler your using and what error your getting?

Azaezel:
vs2013, was spamming
 use of undefined type 'OpenVRTrackedObject' without a remout and
C2065: 'OpenVRTrackedObject' : undeclared identifier with as the first two quick 'fixes'

Areloch:
Sorry for the huge wait time on testing this(had to move all the vive stuff around). I grabbed this, merged with devhead and it compiled on the first try, no problems.

I'll test it compiling in vs2013 next to make sure it's not an oddball compiler version issue.

jamesu:
Rebased to latest dev branch

PR #1691, opened 2016-07-15
GuiObjectView was using an incorrect frustrum when setting up SceneCa…

OTHGMars:
…meraState

### DIFF
--- a/Engine/source/T3D/guiObjectView.cpp
+++ b/Engine/source/T3D/guiObjectView.cpp
@@ -489,16 +489,6 @@ void GuiObjectView::renderWorld( const RectI& updateRect )
 
    // Determine the camera position, and store off render state.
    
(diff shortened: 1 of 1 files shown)


Azaezel:
step in the right direction conceptually anyway looks like. notes:

with a reconfigured game\scripts\client\init.cs of
https://gist.github.com/Azaezel/5635dd0472bd973c8a141b7d4fa7e849#file-gistfile1-txt-L77-L80
vs the stock 
https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/game/scripts/client/init.cs#L109-L115

we get http://imgur.com/a/8TYeB. Now it did kill off the bad flickering on the gl and dx sides that's presently occurring, but mSaveFrustum still does need to be updated looks like.

PR #1693, opened 2016-07-16
bugfix: emissive 

Azaezel:
emissive surfaces  need to use full lighting values, not no lighting values.


andr3wmac:
Emissive means the pixel is a light source. Therefore no lighting should be applied to it. This should be detected in the lighting combine pass and the original color of that pixel should be output without additional lighting calculation. 

The reason your current approach won't work is because lights are additively blended into the lighting buffer, so if you output (1.0, 1.0, 1.0) into the lighting buffer for emissive surfaces than 2 overlapping lights on an emissive surface will add up to (2.0, 2.0, 2.0) in the lighting buffer. This will double the intensity of the emissive pixel which isn't how an emissive pixel should behave.

Issue #1696, opened 2016-07-25
Grenade Launcher shoot from the player's face


John3:
Video bug:
https://youtu.be/FbaRBLfGsj4

PR #1697, opened 2016-07-25
fix issue #1696 

John3:
Video bug:
https://youtu.be/FbaRBLfGsj4

fix issue #1696 
https://youtu.be/F4XybiN4zOA


Areloch:
Rather than removing the first person mesh, wouldn't it be better to muss with the correct muzzle vector settings to try and get it to offset properly?

John3:
Thx @Areloch Ok I will check it out.

PR #1706, opened 2016-07-28
Tweaks the detail textures for the terrain

Areloch:
Tweaks the detail textures for the terrain to work better with the linear pipeline.


Duion:
That is totally unnecessary you can just turn the intensity value down, resaving art assets just adds size to the repo and does nothing useful.
And this may not even be the final setup for the terrain textures so in the next update you would have to change them again and maybe sometime later again etc

PR #1720, opened 2016-08-01
navmesh file load error-fix

Azaezel:
don't try to read 0 length buffers


RichardRanft:
Yup - found it, this is exactly what I did.

PR #1726, opened 2016-08-07
added path @dottools

John3:
this https://github.com/GarageGames/Torque3D/pull/1725#issuecomment-238079691


Lopuska:
it's not really an error, but ok. Inclusions with quote characters are often supposed to be relative.

Areloch:
Yeah, for consistency, that's a fair enough change.

PR #1736, opened 2016-08-10
Allow to create a PS datablock from the editor.

John3:
This fix allow to create a PhysicsShape datablock from the editor.


dottools:
Hate to nitpick, but the code style used in the new line is inconsistent with the rest of the file. Should be like so:
```C#
   if (server)
      return true;
```

Lopuska:
closing as duplicate of better fix.

PR #1737, opened 2016-08-11
just removed a comma from a commented section

SeanTheGuyThatAlwaysHasComputerTrouble:


andr3wmac:
1) You're changing the license text of a third party lib. The same license is in countless files in T3D. Why only one file change?
2) The comma is part of the original MIT license. See here: https://opensource.org/licenses/MIT
3) The comma is valid. Its used to inject additional information that isn't part of the main statement. For instance the sentence without the added 'free of charge' would be: "Permission is hereby granted to any person obtaining a copy.."

Areloch:
Hey Sean, thanks for taking the time to submit a PR!

However, in this case, I'm in agreement with andrew. The license as-is matches to the original license file(which is legally important) and if there WERE to be a change, it'd need to be in all files holding that change.

So I'm going to go ahead and close this one.

PR #1756, opened 2016-08-30
Fix load with DTS shapes introduced with HW skinning changes

jamesu:
Due to a few typos and bad edge cases, meshes which use the following wont load correctly with the HW skinning changes:

- Secondary UV coords
- Colors
- Parented meshes combined with removed meshes
- Loading meshes using a 64bit memory address

This PR fixes this problem.


dottools:
Well done. This is for you :cookie:  ;)

PR #1760, opened 2016-09-02
Update libpng

Bloodknight:


dottools:
Should get it rid of any unnecessary files like `contrib` subdirectory and others. Really only the core library source files and the license file are what should be in the `lpng` directory. No need to keep everything that the official libpng releases provide.

rextimmy:
Can confirm this fixes the crash on exit problem with x64 builds. Might be a bit nicer with above suggestion and also just the one commit. Also version 1.6.25 was released on sept 1st, i have a commit here for it if you want it (removing the contrib files etc too) https://github.com/rextimmy/Torque3D/commit/28337d4cd6533791e83b48772c918446fd2927ef

Bloodknight:
@dottools I copied like for like what was in the original folder, i thought i had already removed most of the unneeded stuff, i hadnt actually considered extra cleanup, made notes for future libs.

@rextimmy i'm more than happy for you to close this and use your newer branch for PR, upgrading the lib  solved my personal problems with the engine and i just shared :)

Lopuska:
Closed as duplicate of #1778

Areloch:
Even though went with timmy's push of the latest version, a huge thanks to ya, Blood, for spearheading getting the lib updated : )

PR #1761, opened 2016-09-02
Intrinsicsfix

Bloodknight:


Areloch:
I'm unsure, but is having it check for either x86 or x64 redundant? It's always going to be one of those two for standard hardware, right? Removing those checks completely strikes me as being the simpler approach if it's always going to pass on one of those.

Otherwise, this looks pretty awesome.

dottools:
@Areloch Possibly for files that are specific to the target architecture, but not for files like `oggTheoraDecoder` where it needs to know what the target platform is. Especially if T3D were to be ported to Android (ARM majority) or other platforms/devices.

JeffProgrammer:
Muh PowerPC support.

PR #1762, opened 2016-09-06
lightbuffer (aka brightness and shadow) always comes last as a mul

Azaezel:


andr3wmac:
This is incorrect. RGB in the lighting buffer represents diffuse lighting. A is the specular. This alone is an inaccurate hack (packing specular into a single channel) but you're making it worse by adding specular then multiplying by diffuse. For instance if you have a material that produced black (0,0,0) for diffuse but still had a specular highlight (like metallics) then you'd add your specular to albedo and then multiply by 0,0,0 and get nothing but black.

Lopuska:
Yup!

PR #1763, opened 2016-09-06
banding: conforms misbehaving postfx to the hdr buffer format

Azaezel:
until we have a proper srgb compatible buffer to minimize that with, we'll have to settle for 16F looks like


andr3wmac:
No criticism of the actual code but your description implies SRGB and banding/16F are related when they aren't. SRGB just handles the linear/gamma correction entirely for you. Banding is related to lack of precision, which is why RGBA16F fixes it. Replacing this with RGBA8_SRGB would not fix banding, it would just remove the need for the toLinear and toGamma calls.

PR #1766, opened 2016-09-12
Replace Epoxy with Glad

JeffProgrammer:
Sorry folks. Looks like epoxy is turning out to be a lemon.

1) epoxy isn't really maintained anymore.
2) epoxy as we know had trouble with VS2015 and we are thus compiling it with optimizations turned off in release.

And now we get a number 3!

3) So, I was adding support back into Torque 3D so that you can compile the engine as a DLL using CMake. It turns out epoxy defines its own DLLMain method, making it impossible to statically link into another DLL unless you modify the library itself. 

Timmy suggested we use Glad, and we both have used it in the past. It is super simple to use. The glad generator is the OpenGL 4.5 core profile + all of the extensions that are usable in the core profile. WGL and GLX are also added.

If we could have a Linux tester be so kind to let me know if the GLX portion is working (do we even need that since we use SDL) that would be great.

As they say, three strikes and you are out...


Lopuska:
As long everything work fine is great!
Someone have time to test it quickly for double check?
Also there's a little conflict to resolve.

JeffProgrammer:
fixed merge conflicts @Areloch @Lopuska

Lopuska:
Awesome. Does someone tested that on Linux?

Areloch:
Looks like the UNIX systems(both lin and mac) have some issues with the cmake config.

I'll have a look into it.

JeffProgrammer:
K we should be good now.

PR #1767, opened 2016-09-12
Add the ability to compile the engine as a DLL.

JeffProgrammer:
You will have to use this PR in conjunction with #1766 if you intend to use OpenGL on windows. The reason is Epoxy has a DLLmain method in it. See the issue listed above for more information.

This PR simply adds an option in cmake called TORQUE_SHARED so that you can compile the engine as a DLL and have a "stub" exe, just like the project generator allowed.

*Do not merge in until 1766 is merged.*


JeffProgrammer:
I should stress that right now this is Windows *only*

JeffProgrammer:
we got it working on mac, need to merge it in.

Issue #1701, opened 2016-07-27
Normal maps on terrain broken


Duion:
It looks like normal maps on terrain are broken, when I add them nothing changes, instead there is ugly graphic artifacts:
http://duion.com/files/temp/screenshot_TG_DirtPit_day001-00001.png

VinnyVicious:
Confirmed on 3.9.

Areloch:
Can you link me to the textures you're using so I have the exact reproduction case?

Duion:
I think any will do, the one I used on the screenshot is this https://github.com/Duion/Uebergame/tree/development/art/terrains/slope
And where I noticed the normal maps on regular shapes were broken was on these models
https://github.com/Duion/Uebergame/tree/master/art/shapes/barriers/inflatable/inflatable_barriers_01

Duion:
I think the feature that made the normal map ambient darker, if you change the ambient light to darker was broken as well, it recently got added from BeamNG I think, but now seems broken again, since the normal maps do look very flat and not as crispy as in 3.8 apart from them being broken with those glitches occasionally.

Duion:
The normal maps are broken in dx9, dx11 and opengl, so it is not just one renderer that causes it.

RichardsGameStudio:
It seems that it is caused (at least in some instances) by the part in:
`Var* GBufferConditionerHLSL::_conditionOutput( Var *unconditionedOutput, MultiLine *meta )
`
where the depth is encoded into 2 channels (line 360-369)

I manually removed the multiplication float2 part from the shader of a material which had this issue, which resulted in a correct rendering of this material (but this certainly is no working method).

Changing this in the engine (line 367) is not an option, this resulted in other unwanted behavior for materials that are correct. The material that I had an issue with contained glowing and subsurface scattering. Don't know if the problem with terrains issue #1740 has a relation with this.

Azaezel:
From what I've been able to determine thusfar it is indeed related to multipass scenarios and gbuffer referencing. terrain with >3 layers for instance having to have a second blendpass, or a model with two layers only the first of which contains a normalmap both wig out in a similar fashion.

Duion:
I fixed it on the model by just adding a normal map into the second UV layer as Azaezel said, but that is not possible with terrain.

RichardsGameStudio:
Confirmed. Adding a normal map into the 2nd UV layer fixes this issue for shapes.

Duion:
[...]

PR #1774, opened 2016-09-15
embeds blendtotal into the low bit for the normal|depth buffer 

Azaezel:
for terrains to support multi-pass blending. 

Do note this will only address https://github.com/GarageGames/Torque3D/issues/1701 . Comprable blending artifacts for normals on objects with multiple layers with the normal maps on layer 0 and subsequent layers lacking a normal will need to be addressed separately.

### DIFF
--- a/Engine/source/terrain/terrCellMaterial.cpp
+++ b/Engine/source/terrain/terrCellMaterial.cpp
@@ -548,8 +548,8 @@ bool TerrainCellMaterial::_createPass( Vector<MaterialInfo*> *materials,
       // MFT_TerrainAdditive feature to lerp the
       // output normal with the previous pass.
       //
-      //if ( prePassMat )
-         //desc.setColorWrites( true, true, false, false );
+      if ( prePassMat )
+         desc.setColorWrites( true, true, true, false );
    }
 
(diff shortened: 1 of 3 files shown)


Engine/source/terrain/terrCellMaterial.cpp
```diff
@@ -548,8 +548,8 @@ bool TerrainCellMaterial::_createPass( Vector<MaterialInfo*> *materials,
       // MFT_TerrainAdditive feature to lerp the
       // output normal with the previous pass.
       //
-      //if ( prePassMat )
-         //desc.setColorWrites( true, true, false, false );
+      if ( prePassMat )
+         desc.setColorWrites( true, true, true, false );
```
Lopuska: Does color writes on prepass target was the responsible of the green shift we had long time ago on decals while working on fully deferred branch?
(Just to be sure before merge)

Azaezel: yes. why we keep the blue channel around now regardless.

Lopuska: great!

PR #1778, opened 2016-09-17
Libpng update to 1.6.25 (fixes x64 crashing on exit)

rextimmy:
This is the libpng update with the latest version, fixes a crash on exit with 64bit systems and i also removed the unnecessary stuff we don't need that ships with libpng.


Lopuska:
Thanks! It was the same fix I did long time ago to resolve the frustrating exit crash, though I never merged it :D

PR #1785, opened 2016-09-28
MacOS platform support

rextimmy:
Support for 64bit macOS platform. Minimum requirement is macOS 10.9.

Port originally based off the excellent work done by @JeffProgrammer and mixed in with some code from out little brother Torque2D.

*Retina displays are currently unsupported as i didn't have access to retina display and i'm not keen on committing untested code. The retina display will work, however the resolutions will be incorrect.  SDL2 does have retina support, so it's just a matter of finding a user with a retina display who can test some code for us before committing it.


JeffProgrammer:
I will test on a 2011 mac mini with an AMD 6000 series graphics card, and a 2013 retina macbook pro with Intel Iris Pro.

JeffProgrammer:
Tested on Iris Pro. Confirm working on MacOS Sierra, compiled with XCode 7.0.1 and used the 10.11 SDK, target 10.9 x64.

Lopuska:
Thanks to every contributor to make this real 👍

Areloch:
My results on an iris confirm Jeff's. Compiled in capitan and sierra without issue, ran on both.

PR #1800, opened 2016-10-14
clang reports: unclear || + && and &+| mixes.

Azaezel:


Azaezel:
I'll urge folks to take extra care validating this one in particular. While this run largely preserved present functionality, it did correct a couple very subtle bugs. Any oversights on my part are likely to be just as subtle.

PR #1817, opened 2016-10-25
Preliminary IPV6 Support

jamesu:
Implements preliminary IPV6 support. 

What works:
- Connecting to ipv4 addresses (yes, this is backwards compatible)
- Connecting to ipv6 addresses
- Browsing for ipv6 servers on a LAN
- Working with ipv6 TCP sockets
- Returning IPV6 addresses from the master server
- Registering an IPV6 server with the master server

IPv6 addresses may be represented by the following formats:
- `IP6:address`
- `IP6:[address]:port`
- `IP^:[address]:port%interface`

_NOTE_: ipv6 by default is disabled. You need to set `$pref::Net::IPV6Enabled` to 1 in order for a server to listen on ipv6 and also to be able to query ipv6 servers.

For a suitable master server to test with, check out https://github.com/jamesu/PushButton-Master-Server/tree/ipv6_support

For LAN testing, any ipv6 network should work - just make sure everything is in the same subnet, etc.

This particular code has been tested on windows and linux, though should work on macos too.

Note that this PR also changes the NetSocket to a struct, as a lot of code was liberally referring to sockets using U32. This is no longer a safe assumption, especially on platforms such as WIN64. This PR consequently resolves a potential bug if windows returns a socket handle which cannot be stored in U32.

### DIFF
--- a/Engine/source/app/game.cpp
+++ b/Engine/source/app/game.cpp
@@ -162,6 +162,13 @@ DefineConsoleFunction( setNetPort, bool, (int port, bool bind), (true), "(int po
    return Net::openPort((S32)port, bind);
 }
 
+DefineConsoleFunction(isAddressTypeAvailable, bool, (int addressType), , "(protocol id)"
+	"@brief Determines if a specified address type can be reached.\n\n"
+	"@ingroup Networking")
+{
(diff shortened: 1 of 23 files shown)


Azaezel:
need
 #elif defined ( TORQUE_OS_MAC )
...
 #include <net/if.h>

Issue #1707, opened 2016-07-28
Shadow lags behind with sun object


Duion:
When I use the basic sun object instead of scattersky, which I have to do in a level where I use a skybox, then the shadow of my player extremely lags behind, it feels like the game is just running at 5 fps according to the shadow, but the rest runs fine.

John3:
If i'm not wrong, from inspector panel, you need to change the shadow refresh

Duion:
Related:  http://forums.torque3d.org/viewtopic.php?f=9&t=387
Yes I know now where it comes from, I just wondered why scatterSky is smooth while sun is laggy.

Duion:
It seems to be just an issue with the default settings the objects start out with, so you want me to close this issue?

Duion:
I found another issue related to that, now I have setup static and dynamic shadow objects, when I change the refresh rate from the static objects it works, but when I change the refresh rate of the dynamic objects it always stays smooth, no matter how high the value is I enter there.
So it seems like you cannot adjust the refresh rate of dynamic objects.

Azaezel:
https://github.com/GarageGames/Torque3D/pull/1640 had to shut that end off till we sorted out why it was causing translucent materials to flicker.

Areloch:
It may be worth it to just hide the field in the meantime from appearing in the inspector, since it is currently disabled functionality.

Just to avoid confusion.

Azaezel:
fair enough.

Duion:
I get the idea of having the shadows update less often to save performance, but I'm still trying to find good numbers, the default 250ms are too much, I use 128ms, so roughly half, but you can still see the shadows popping, if you move your mouse fast. For optimal results for fast paced gameplay you will probably have to go below 100ms.

PR #1824, opened 2016-11-10
Hides the light's dynamic refresh rate field in the editor

Areloch:
Hides the light's dynamic refresh rate field to avoid confusion, leaving only the static refresh rate field to be edited in reference to #1707

### DIFF
--- a/Engine/source/T3D/lightBase.cpp
+++ b/Engine/source/T3D/lightBase.cpp
@@ -93,7 +93,7 @@ void LightBase::initPersistFields()
       addField( "brightness", TypeF32, Offset( mBrightness, LightBase ), "Adjusts the lights power, 0 being off completely." );      
       addField( "castShadows", TypeBool, Offset( mCastShadows, LightBase ), "Enables/disabled shadow casts by this light." );
       addField( "staticRefreshFreq", TypeS32, Offset( mStaticRefreshFreq, LightBase ), "static shadow refresh rate (milliseconds)" );
-      addField( "dynamicRefreshFreq", TypeS32, Offset( mDynamicRefreshFreq, LightBase ), "dynamic shadow refresh rate (milliseconds)" );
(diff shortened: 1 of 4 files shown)


JeffProgrammer:
i assume this is because dynamic is just updated every frame so no need to change?

Areloch:
Yep.

JeffProgrammer:
makes sense to me

Duion:
Why hide it? You can also change it for performance benefit, according to my tests you can increase it from 8ms to 15ms without being noticed, this will update it every frame on 60fps base and gives decent performance benefit compared to the 8ms default.

Duion:
I don't really know if it makes sense to have those fields at all for each individual mission, since it is more a global variable.

Areloch:
It can vary by mission. For example, one mission where you have the scatter sky and time of day, you'd want a lower refresh rate so you don't get major jitters, and the another map where the time doesn't change, so you can have a longer refresh rate for some performance gain.

Duion:
You need constant refresh rate, since when you turn fast or respawn you see the shadows popping in anyway.
Time of day usually moves so slow, it does not need faster refresh rate.
I see it that either having both fields visible makes sense so you can adapt everything to your needs in each mission or hide both with good default values, that can be individually adapted for the whole game.

Azaezel:
Per-Light values are there for fast moving or animated lights. Slowish refresh rates there would cause shadows to be 'drug along' when in motion.

Duion:
The shadows also "drug along" if you yourself are moving, for example turning fast with your mouse view or respawning into the game, I can see shadows popping with the static update rate as low as 50ms, only below that it becomes invisible.

Azaezel:
yeah, https://github.com/Azaezel/Torque3D/tree/shadowCacheCorrection should have the fix for respawning. still eyeballing how best to handle quick turning.

Areloch:
Yeah, I'm taking a look at an angle-test function so we can toss that into the mix as az said. Should help cut back on shadows lagging when you spin around really fast.

Areloch:
I'll go ahead and merge this for now, and we can consider the caching correction and maybe bumping the dynamic refresh rate up as separate PRs/issues for organizational purposes.

Whenever we make more proper utilization out of adjusting the dynamic refresh rate, we can always  un-hide the field from the editor again.

PR #1830, opened 2016-11-24
Fix  for NavPath is not updated when navMesh has change. 

pacomont:
It's not working because on onNavMeshUpdate the string comparation will never succeed, and will never call  call navpath::plan()

```
DefineEngineMethod(NavPath, onNavMeshUpdate, void, (const char *data),,
   "@brief Callback when this path's NavMesh is loaded or rebuilt.")
{
   if(object->mMesh && !dStrcmp(data, object->mMesh->getIdString()))
      object->plan();
}
```

### DIFF
--- a/Engine/source/navigation/navMesh.cpp
+++ b/Engine/source/navigation/navMesh.cpp
@@ -838,7 +838,7 @@ void NavMesh::buildNextTile()
          ctx->stopTimer(RC_TIMER_TOTAL);
          if(getEventManager())
          {
-            String str = String::ToString("%d %.3f", getId(), ctx->getAccumulatedTime(RC_TIMER_TOTAL) / 1000.0f);
+            String str = String::ToString("%d", getId());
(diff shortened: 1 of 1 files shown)


crabmusket:
Good catch. I believe I put the timing information in the callback so that the UI would be able to display the data. See [here](https://github.com/GarageGames/Torque3D/blob/c83efa59e0a11fc15c68afaf106807643191ac6f/Templates/Empty/game/tools/navEditor/NavEditorConsoleDlg.gui#L136).

So the **fastest** solution would be to change the code that you've copied in the issue, to take the two-word string into account.

Maybe the **better** solution would be to emit a separate timing event and make the UI listen to that.

Areloch:
Yeah, having a separate timing even would be a good thing to add in. Probably can just make an issue for that and will merge this fix in.

pacomont:
I'm having a peak workload those days. Let me few days and I could give a try.

PR #1835, opened 2016-11-29
Adds some helpful utility math functions.

Areloch:


crabmusket:
Might be nice to mention the random distribution used (assuming uniform?).

Areloch:
Fair point. it just uses the straight mRandomLCG, so yeah, just seeds off deadbeef.

PR #1839, opened 2016-12-01
directional coloration for pathnodes,

Azaezel:
as well as enlarges both the look and 'physical' handle size for ease of use.


rextimmy:
Working ok here, i like it.

PR #1831, opened 2016-11-26
fix ColladaExporter

John3:
Remove old colladamax format, fixed up bad coordinates, added new collada format, valid against the COLLADA 1.4.1 schema, added to load textures and UVs.

Test it! [ T3D_ColladaExporter (15MB)](https://drive.google.com/file/d/0B9GUwtDHCp_eYlpuZHQ1aEhGRmM) Thanks!

Open Issue: https://github.com/GarageGames/Torque3D/issues/467

PR #1843, opened 2016-12-03
Update version.h

rasteron:


Duion:
The next version that will come out now is 3.10, no idea how you write that in the current version number system...

rasteron:
It is still the current version isn't it? which is 3.9? or 3.10?

https://github.com/GarageGames/Torque3D/releases/tag/3.9

rasteron:
..and question, how do you go to 3.10 from 3.9? Is it supposed to be 4.0?

Areloch:
Haha, yeah, it can seem a bit weird for the number convention at first.

I'll go ahead and merge this now, because it's correct, and just remember to increment it myself to 3.10 when we push the 3.10 release. :)

rasteron:
Ah ok thanks Jeff :)

crabmusket:
> no idea how you write that in the current version number system...

That's a very good question :p. Depends on how that number is used. Could make it 40000 while maintaining monotonic increase, if its role is to be compared numerically to other version numbers.

Duion:
You could change the engine version instead and keep the regular version number, but this may be confusing as well since most people don't know there is an engine versioning system as well.

I would not make it 31000, I would rather make it 3910 or so, but engine version change may make sense, since there is a major shift in technology with the move to PBR.
As a bonus you could invent a new name for Torque3D, here are the current names:
https://github.com/rasteron/Torque3D/blob/a3f9365347134eecae4d3a1a30a90cf65818c39b/Engine/source/app/version.h#L30-L36
I already named my engine fork:
https://github.com/Duion/Torque3D/blob/ueberengine-master/Engine/source/app/version.h#L37

We could name the new Torque "Torque 10"

John3:
This is a minor issue, AFAIK v3.10 was something kinda "unexpected", so maybe this will not gonna happen again.

BTW I did this change in this PR https://github.com/GarageGames/Torque3D/pull/1831 because I need it for the collada exporter, so the v3.10 will not forgotten.

PR #1852, opened 2016-12-10
Adds some bake-to-collada functions

Areloch:
Makes it so prefabs can correctly 'export to collada' by running the buildPolyList function on it's children.

Also adds a 'Bake Selection to Mesh' option to the tools menu in the editor to export the selected mesh to a collada file, and then replaces the selection with a TSStatic.


dottools:
But where's `bakeChristmasCookies()` ? :(

JeffProgrammer:
I'll bake you some cookies @dottools :)

PR #1857, opened 2016-12-13
Add Filter for dyamic objects when building a navmesh in the editor

Areloch:
Makes it so dynamic objects don't contribute to the navmesh when building in the editor. This avoids player objects, vehicles, etc from blocking parts of the navMesh.

When in-game and tiles are rebuilt, it keeps the contribution of dynamic objects so AI can path around other players, etc.


Areloch:
This logic can be handled better.

Issue #1863, opened 2016-12-15
OpenGL console warning


Duion:
I get this entry in the console when running in OpenGL multiple times:

GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.
GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.
GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.
GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.
GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.
GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.
GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.
GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.
GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.
GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.
GFXCardProfiler (OpenGL) - Unknown capability 'GL::Workaround::noCompressedNPoTTextures'.

Any ideas what this is?

PR #1868, opened 2016-12-15
GL::Workaround::noCompressedNPoTTextures profile is no longer used

rextimmy:
fixes up #1863


JeffProgrammer:
This will be fine. We no longer have to check for non power of 2 textures, since that is a legacy OpenGL thing.

PR #1876, opened 2016-12-20
Fixes Bullet not supporting holes in terrain.

Areloch:
Fix from @Lopuska 's PhysicsGMKRefactor branch!


rextimmy:
Tested ok here. Thanks @Lopuska

Lopuska:
The fix follows the exact recommendations directly from Erwin Coumans on his bullet forum. Bullet is a kind of library designed to extend its feature set, so some features need to be added by redefining base classes. That's the reason holes were only implemented on PhysX, because officially bullet doesn't support heightfield holes.

Areloch:
@Lopuska ah, that's pretty interesting. Thanks for the hard work!

PR #1877, opened 2016-12-20
Default port fix

jamesu:
Resolves two things:

- Incorrect default port used for clients
- Incorrect multicast address print


Azaezel:
verified

Issue #1746, opened 2016-08-25
SDL2 app icon missing


Duion:
After compiling the engine with the flag "TORQUE_SDL" so that it uses the new SDL2 I noticed that the app icon in the upper left of the window is missing and it shows a default icon instead, but otherwise the icon shows fine, like next to the executable or in desktop shortcuts.

wshearn:
I have found a solution(tested on Linux but should work on Windows).

Follow the image export here and save it as My Projects/<Project>/source/icon.h: https://caedesnotes.wordpress.com/2015/04/13/how-to-integrate-your-sdl2-window-icon-or-any-image-into-your-executable/

When exporting I kept the name gimp_image.

Apply the following

``` diff
index b1d3d04..1d244b6 100644
--- a/Engine/source/windowManager/sdl/sdlWindowMgr.cpp
+++ b/Engine/source/windowManager/sdl/sdlWindowMgr.cpp
@@ -26,6 +26,9 @@
 #include "core/strings/unicode.h"
 
 #include "SDL.h"
+#if TORQUE_SDL_ICON
+#include "icon.h"
+#endif
 
 // ------------------------------------------------------------------------
 
@@ -163,6 +166,28 @@ PlatformWindow *PlatformWindowManagerSDL::createWindow(GFXDevice *device, const
    window->mOwningManager = this;
    mWindowMap[ window->mWindowId ] = window;
 
+#if TORQUE_SDL_ICON
+   Uint32 rmask, gmask, bmask, amask;
+#if SDL_BYTEORDER == SDL_BIG_ENDIAN
+   int shift = (gimp_image.bytes_per_pixel == 3) ? 8 : 0;
+   rmask = 0xff000000 >> shift;
+   gmask = 0x00ff0000 >> shift;
+   bmask = 0x0000ff00 >> shift;
+   amask = 0x000000ff >> shift;
+#else // little endian, like x86
+   rmask = 0x000000ff;
+   gmask = 0x0000ff00;
+   bmask = 0x00ff0000;
+   amask = (gimp_image.bytes_per_pixel == 3) ? 0 : 0xff000000;
+#endif
+
+   SDL_Surface* icon = SDL_CreateRGBSurfaceFrom((void*)gimp_image.pixel_data, gimp_image.width,
+   gimp_image.height, gimp_image.bytes_per_pixel*8, gimp_image.bytes_per_pixel*gimp_image.width,
+   rmask, gmask, bmask, amask);
[...]

PR #1878, opened 2016-12-22
Fixes window icons with SDL

Areloch:
Fixes window icons with SDL, hooking it through the var $Core::windowIcon as the path.

Also adjusted the splash window icon to use the var $Core::splashWindowImage for it's path.

In reference to #1746


Duion:
Tested and it works.

PR #1883, opened 2016-12-23
DX11 updates

rextimmy:
Various updates i been meaning to chuck in for ages:
* Caching sampler states
* Better tracking of sampler ranges in use for binding time
* Feature level 10 & 10.1 support
* Full-screen fix where window has no focus on startup
* Don't add DX11 adapters to the adapter list when they are not supported by the system (causes a crash when selected)
* D3DPER debug events


RichardsGameStudio:
Nice!

PR #1891, opened 2016-12-30
D3D11 shadermodel version fix

rextimmy:
*Fixes up a problem where shader model version was not getting set for shader macros.
*Adds feature level string to the version string of the card profiler.


Lopuska:
👌👍

PR #1894, opened 2017-01-02
enable video recording

John3:
EDIT: sry I forgot to add a description..... this add a better function of the video recording and add a shortcut.


Areloch:
When testing this, I ran into an issue where theora would crash in debug mode. Seemed to work fine in release, and it's not really an issue with this PR itself, so much as some oddball bug in theora. So the main question would be, should we just disable theora recording in debug?

I mean, recording already has a performance hit, and compounding that with the perf hit in debug seems like it'd be a poor combination even if there wasn't some issue in theora itself, so would it make sense for us to block theora from recording when in debug just to head off issues?

Outside the above thing, this PR seems to work perfectly well in release, so if we can decide on what to do about the debug crash, we can get this sorted out quick.

PR #1900, opened 2017-01-04
Added _physicsEnabled() function to Player class

rextimmy:
This PR allows you to move the player around in the editor with a physics plugin when you don't have the simulation enabled (default for the editor)

### DIFF
--- a/Engine/source/T3D/player.cpp
+++ b/Engine/source/T3D/player.cpp
@@ -51,6 +51,7 @@
 #include "T3D/fx/cameraFXMgr.h"
 #include "T3D/fx/splash.h"
 #include "T3D/tsStatic.h"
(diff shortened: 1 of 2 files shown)


Engine/source/T3D/player.cpp
```diff
@@ -4994,6 +4995,18 @@ void Player::_handleCollision( const Collision &collision )
       queueCollision( collision.object, mVelocity - collision.object->getVelocity() );
 }
 
+bool Player::_physicsEnabled()
```
dottools: This looks like something very expensive to do considering that it'll be called multiple times per Player object instance per tick time.

rextimmy:
Yeah you are right, i'll delete the pr

PR #1904, opened 2017-01-05
Fix net tests

elfprince13:
There was a silly error where a bunch of Net::sm* stuff was being declared as standalone static variables rather than as static members.


dottools:
Formatting tabs vs spaces in this PR too. :/

elfprince13:
Should be all set.

PR #1907, opened 2017-01-05
OSX de-deprecation (profiler and macFileIO)

elfprince13:
Fixed a number of deprecated API usages on OS X, as well as made the move and copy operations share a bit more code, and both a bit safer.


dottools:
Formatting problem with tabs vs spaces.

elfprince13:
Is anyone aware of a simple setting to force Xcode to use the style conventions T3D on a per-project basis? Currently it auto-corrects spaces to tabs when I edit a line, which makes this a bit miserable to match up.

elfprince13:
Okay - should be good here now with spaces.

JeffProgrammer:
http://editorconfig.org/ @elfprince13

PR #1915, opened 2017-01-06
Fixed semaphore struct to class

elfprince13:
platformSDL had `PlatformSemaphore` defined as a struct rather than a class, which conflicted with the original declaration.


Lopuska:
Thanks (:

PR #1916, opened 2017-01-06
SFX Variance Overflow

elfprince13:
The compiler warned that `mVariance[1]` was out of bounds. This should have been varying the index along the second axis of mVariance, and keeping the first axis fixed at 0. The compiler didn't catch that there was also a type error, because `Con::printf` isn't typesafe.


Lopuska:
👍

PR #1923, opened 2017-01-12
String table empty string

elfprince13:
I was going to merge some old changes of mine from the 1.0.1 days for having a global "empty stringtable entry", and realized the engine had since added the functionality, but didn't appear to actually be using it anywhere. This PR is a global find and replace with `StringTable->insert("")` -> `StringTable->EmptyString()`.


crabmusket:
Whitespace :cry:

elfprince13:
```
[filter "tabspace"]
	clean = expand -t 3
	smudge = unexpand -t 3
```

Is a useful thing to prevent my editor from putting in tabs where I don't want them. The downside is...it's a little aggressive about making sure *every* file complies, and apparently the codebase isn't super consistent.

Areloch:
First) nice PR
Second) I wouldn't fret super hard on the whitespace cleanup part of it. It looks like it conforms stuff to the 3 space, even if parts aren't especially relevent to this PR in particular. We're planning on doing a whitespace cleanup pass for one of the first parts of the push towards 4.0 when 3.10 goes out, along with guideline compliance and the like.

So yeah, wouldn't fret much on it if you inadvertently catch some whitespace cleanup in the course of things. We'll get it all cleaned and consistent one way or another pushing on towards 4.0 :)

elfprince13:
Awesome :)

Areloch:
Looks like a couple conflicts snuck their way in. Can you double-check those so I can get this merged in? Thanks :)

elfprince13:
Should be all good. It was all `NULL` -> `nullAsType<>()` conversions from #1920

PR #1925, opened 2017-01-13
Sdl update fix

elfprince13:
03c2ce46c260852a60731b2311ca1d668760f3a3 appears to have been an incomplete update of SDL 2.0.5 and it broke the build on OS X.


Areloch:
Thanks for this. Thought I had got it tested on mac, but maybe there were merge shenanigans that cause me to mis-test on this end. I'll get this checked out.

elfprince13:
I didn't notice right away after I updated because I had it set to only rebuild the application binary for my project in my IDE, instead of on ALL_BUILD, and it kept happily linking to the old version, so I wouldn't be surprised if something similar was going on for you and others.

PR #1937, opened 2017-01-31
fix warningFlashes() of lightning class

John3:
This PR fix the warningFlashes() method, that doesn't randomly select the spot to strike, all the lightning bolts will always strike in the same place.

EDIT: typo :D


Engine/source/T3D/fx/lightning.cpp
```diff
@@ -742,6 +743,9 @@ void Lightning::warningFlashes()
       {
          LightningStrikeEvent* pEvent = new LightningStrikeEvent;
          pEvent->mLightning = this;
+
```
dottools: tabs vs spaces issue. :P

PR #1938, opened 2017-02-01
added strikeObject lightning feature

John3:
added new feature strikeObject to lightning.

Now you can throw a lightning to objects  derived from shapebase, like players, Items, AI, etc.

EDIT: typo

### DIFF
--- a/Engine/source/T3D/fx/lightning.cpp
+++ b/Engine/source/T3D/fx/lightning.cpp
@@ -870,25 +870,30 @@ void Lightning::strikeObject(ShapeBase* targetObj)
    Point3F strikePoint = targetObj->getPosition();
    Point3F objectCenter;
 
-   targetObj->getObjBox().getCenter(&objectCenter);
-   objectCenter.convolve(targetObj->getScale());
-   targetObj->getTransform().mulP(objectCenter);
+   Box3F wb = getWorldBox();
+   if (!wb.isContained(strikePoint))
+        return;
+
+   Point3F targetRel = strikePoint - getPosition();
+   Point3F length(wb.len_x() / 2.0f, wb.len_y() / 2.0f, wb.len_z() / 2.0f);
+
+   Point3F strikePos = targetRel / length;
 
    bool playerInWarmup = false;
    Player *playerObj = dynamic_cast< Player * >(targetObj);
    if (playerObj)
    {
-	   if (!playerObj->getControllingClient())
-	   {
-		   playerInWarmup = true;
-	   }
+       if (!playerObj->getControllingClient())
+       {
+           playerInWarmup = true;
+       }
    }
 
    if (!playerInWarmup)
    {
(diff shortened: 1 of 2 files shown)


Areloch:
Noted 2 things with this:

1) I hadn't initially realized strikeObject only works with shapeBase. I tried passing in a static before and it crashed. We should probably make it either SceneObject, or if we really do only want ShapeBase objects for targets, add a safety check in there if the wrong type is passed in.

2) The way the strikePoint is passed along is a 0-1 range representing the lighting object's world box, to keep the net events as compact as possible. When you're getting the strike point based on the target object's world transform, the numbers are too big for the expected bit size and it breaks. You'll want to convert the targetObject's world transform to be the local offset of the lighting's world box. That would fix that issue.

If these can get sorted, shouldn't be any issues rolling this in

John3:
Hey Areloch oops sorry for not added the safety check... Well all ShapeBase objects can receive damage, TSSTatic do not have damage handling, that's why I use ShapeBase. :D :D 

I will check the strikepoint to net events. thks

dottools:
This PR also has tabs vs spaces indention conflict issue.

RichardsGameStudio:
Thanks for the setup. If you change things like below, everything should work fine and strikepoints should be 0-1.

```
void Lightning::strikeObject(ShapeBase* targetObj)
{
   AssertFatal(isServerObject(), "Error, client objects may not initiate lightning!");

   Point3F strikePoint = targetObj->getPosition();
   Point3F objectCenter;
   Box3F wb = getWorldBox();
   if (!wb.isContained(strikePoint))
	   return;

   Point3F targetRel = strikePoint - getPosition();
   Point3F length(wb.len_x() / 2.0f, wb.len_y() / 2.0f, wb.len_z() / 2.0f);

   Point3F strikePos = targetRel / length;

   bool playerInWarmup = false;
   Player *playerObj = dynamic_cast< Player * >(targetObj);
   if (playerObj)
   {
	   if (!playerObj->getControllingClient())
	   {
		   playerInWarmup = true;
	   }
   }

   if (!playerInWarmup)
   {
	   applyDamage_callback(targetObj->getWorldSphere().center, VectorF(0.0, 0.0, 1.0), targetObj);
   }

   SimGroup* pClientGroup = Sim::getClientGroup();
   for (SimGroup::iterator itr = pClientGroup->begin(); itr != pClientGroup->end(); itr++) {
	   NetConnection* nc = static_cast<NetConnection*>(*itr);
	   if (nc != NULL)
	   {
		   LightningStrikeEvent* pEvent = new LightningStrikeEvent;
		   pEvent->mLightning = this;

		   pEvent->mStart.x = strikePos.x;
		   pEvent->mStart.y = strikePos.y;
		   pEvent->mTarget = targetObj;

		   nc->postNetEvent(pEvent);
	   }
   }
}
```

John3:
thx richard, I added your suggestion, thax nathan added tab vs space fix.

PR #1950, opened 2017-02-21
Fix for dae2dts tool

SimonBoorer:
This fixes the dae2dts tool when compiled using VisualStudio 2012.

The first compilation issue is that the compiler cannot open `nfd.h`. This is because  `${libDir}/nativeFileDialogs/include`  is not added to the project include paths when `TORQUE_SDL` is disabled. I fixed this by moving `#include <nfd.h>`  to within the  `#if defined(TORQUE_SDL)` preprocessor directive.

The second compilation issue is that the compiler cannot open `al/al.h`. This is because `openal/win32` is added to the project include paths instead of `openal-soft/include`. I fixed this by updating the include paths in the core and T3D project generator modules.

The last compilation issue is that `Component::setOwner(class Entity *)`  is undefined. This is because the `T3D/components` engine source directory is not included. I fixed this by adding it to the T3D project generator module.

After getting the program to compile it then generates a fatal error `SimObject::object missing call to SimObject::onRemove` when run without specifying a DAE file. This is because the program terminates without shutting down the platform. I fixed this by shutting down the platform before terminating.

When you run the program again with a DAE file, it generates an error when `TSShapeLoader` initializes the `TSShape`, as there is no GFX device to initialize the vertex buffers. I fixed this by setting `TSShape::smInitOnRead` to `false` and making `TSShapeLoader` adhere to this setting when initializing the `TSShape`.

After fixing this the program converts DAE models to DTS shapes successfully, however generates a write access violation error when the program exits. This is because the TSShape resource is not unloaded until it goes out of scope, after the platform has shutdown. I fixed this by adding a block scope to the TSShape resource to ensure it goes out of scope before the platform is shutdown.

To generate the project files I am doing the following:
* Open the dae2dts directory
* Run generateProjects.bat
* Open the buildFiles directory
* Copy torque.ico and Torque.rc from VisualStudio 2008\projects to VisualStudio 2012\projects
* Open VisualStudio 2012\dae2dts.sln
* On the dae2dts project, go to it's properties, then:
  * Change Configuration Properties -> General -> Configuration Type to Application (.exe)
  * Change Configuration Properties -> Linker -> General -> Output File to $(OutDir)dae2dts_DEBUG.exe
  *  Remove Configuration Properties -> Linker -> Input -> Module Definition File

I am not sure if this is the correct way to fix all of these issues so please let me know if you have any feedback.

### DIFF
--- a/Engine/source/platform/nativeDialogs/fileDialog.cpp
+++ b/Engine/source/platform/nativeDialogs/fileDialog.cpp
@@ -29,11 +29,13 @@
 #include "console/consoleTypes.h"
 #include "platform/profiler.h"
 #include "console/engineAPI.h"
-#include <nfd.h>
 #include "core/strings/stringUnit.h"
 #include "core/frameAllocator.h"
 
 #if defined(TORQUE_SDL)
+
+#include <nfd.h>
+
 //-----------------------------------------------------------------------------
 // PlatformFileDlgData Implementation
 //-----------------------------------------------------------------------------
--- a/Engine/source/ts/loader/tsShapeLoader.cpp
+++ b/Engine/source/ts/loader/tsShapeLoader.cpp
@@ -1228,7 +1228,11 @@ void TSShapeLoader::install()
    shape->radius = (shape->bounds.maxExtents - shape->center).len();
    shape->tubeRadius = shape->radius;
 
-   shape->init();
+   if (TSShape::smInitOnRead)
+   {
+      shape->init();
+   }
+
    shape->finalizeEditable();
 }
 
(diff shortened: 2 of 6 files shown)


SimonBoorer:
I also had to modify `TSShapeLoader` to prevent clearing the `TSShape` split vertex lists when `TSShape::smInitOnRead` is set to `false`, as the split vertex lists are not being stored in a vertex buffer and are still required.

I tested this using [teapot.dae](https://raw.githubusercontent.com/cathyatseneca/c3dl/master/tutorials/tutorial8/teapot.dae) and the `--compat`  command line argument and verified the model loaded correctly in Torque ShowTool Pro.
teapot

SimonBoorer:
Looks like this solution only works when writing the shape in the old format because the new format depends on the vertex buffers and generates a "vertex size mismatch" error when loaded. Not really sure how to go about solving this.

Issue #1951, opened 2017-02-21
TCPObject is broken


pacomont:
Client connection to a server is not performed. It could be tested trying to connect to a TelnetDebugger service ( startDebugger() ) witch uses TCPObjects. 

When trying to connect **Net::openConnectTo()** is called, but It exists with handleFd = NetSocket::INVALID;

/// To use the debugger, use dbgSetParameters(port, password); in the console
/// of the server to enable debugger connections.  Then on some other system,
/// start up the app (you don't have to start a game or connect to the
/// server) and exec("common/debugger/debugger.cs"); in the console.  Then use
/// the debugger GUI to connect to the server with the right port and password.

Azaezel:
Can you elaborate further on your reproduction case? Masterserver connectivity and lan both seem functional at present. (Will have to wait a bit for reproduction testing on external network testing this end, I'm afraid.)

pacomont:
Of course, sorry... 
[1] Launch 2 instances of the application.
[2] In the first instance (server) execute in console: **dbgSetParameters(3333,"aa");**
[3] In the second instance (client) execute in console: **startDebugger();**
[4] In second instance press the connect button and complete with the address **"127.0.0.1", port "3333" and password "aa"**...
You'll see that nothing happens, the client will not connect with the server.   

You could trace the execution in the client until TCPObject::connect(), and Net::openConnectTo(address); will return NetSocket::INVALID.
`void TCPObject::connect(const char *address)
{
   NetSocket newTag = Net::openConnectTo(address);
   addToTable(newTag);
}`

Image of 2 instances:
[https://goo.gl/SQhUsj](url)

Azaezel:
Yeah, ip connectivity did get tweaked a bit on the backend to support IPV6: https://github.com/GarageGames/Torque3D/pull/1817 likely an oversight... Gotta say, first I'd even heard of that subsystem, so it'll take some digging if nothing springs out at you (since you're more familiar with that subsystem)

pacomont:
Oh wait, I found the real problem (but not the solution, yet...). Forget previous messages, here is the new case:

Using HTTPObject with the stock example from source code ([link](https://gist.github.com/pacomont/961a32674d638eb000858c5c3107442f)), 

I found that using:
`%feed.get("www.google.com:80", "/ig/api", "weather=Las-Vegas,US"); `
it **works** 

but expecifing ip address directly, it **doesn't work**: 
[...]

PR #1952, opened 2017-02-22
Fix issue #1951 tcp object is broken 3.10

pacomont:
Proposed fix for Issue #1951. Checked with IP4. IP6 should work as well but not checked yet.
I'm still not familiar with that technology to perform testings.
Add httpExample.cs to the game directory...
Write in console:
exec("httpExample.cs");
testHTTPObject1();
testHTTPObject2();

Should work the same way both commands.

```
function testHTTPObject1()
{
   WeatherFeed.delete(); // delete previous instances...
   
   // Create the HTTPObject
   %feed = new HTTPObject(WeatherFeed);

   // Define a dynamic field to store the last connection state
   %feed.lastState = "None";

   // Send the GET command
  %feed.get("httpbin.org:80", "/html", "");      

   // check executing in console: echo(WeatherFeed.buffer);
}

function testHTTPObject2()
{
   WeatherFeed.delete(); // delete previous instances...
   
   // Create the HTTPObject
   %feed = new HTTPObject(WeatherFeed);

   // Define a dynamic field to store the last connection state
   %feed.lastState = "None";

   // Send the GET command
   %feed.get("23.22.14.18:80", "/html", "");   
   
   // check executing in console: echo(WeatherFeed.buffer);
}

```
File to perform testings:
[httpExample.zip](https://github.com/GarageGames/Torque3D/files/794697/httpExample.zip)


Engine/source/platform/platformNet.cpp
```diff
@@ -823,13 +823,16 @@ NetSocket Net::openConnectTo(const char *addressString)
       if (socketFd != InvalidSocketHandle)
       {
          setBlocking(handleFd, false);
-         if (::connect(socketFd, (struct sockaddr *)&ipAddr, sizeof(ipAddr)) == -1 &&
-            errno != EINPROGRESS)
+         if (::connect(socketFd, (struct sockaddr *)&ipAddr, sizeof(ipAddr)) == -1)
```
jamesu: From what i've read, there are two possible states here for non-blocking sockets:

returns 0 (e.g. for localhost)
returns -1, errno is EINPROGRESS (e.g. for remote connections)

Probably should be used in this case. Naturally though this seems to be sporadically documented so I cant be 100% sure this is true for every single sockets impl....

Engine/source/platform/platformNet.cpp
```diff
@@ -849,14 +852,20 @@ NetSocket Net::openConnectTo(const char *addressString)
...
-      if (::connect(socketFd, (struct sockaddr *)&ipAddr6, sizeof(ipAddr6)) == -1 &&
-         errno != EINPROGRESS)
+      if (socketFd != InvalidSocketHandle)
       {
          setBlocking(handleFd, false);
-         Con::errorf("Error connecting %s: %s",
-            addressString, strerror(errno));
-         closeSocket(handleFd);
-         handleFd = NetSocket::INVALID;
+         if (::connect(socketFd, (struct sockaddr *)&ipAddr6, sizeof(ipAddr6)) == -1)
```
jamesu: ditto here

Engine/source/platform/platformNet.cpp
```diff
@@ -809,7 +809,7 @@ NetSocket Net::openConnectTo(const char *addressString)
       error = Net::WrongProtocolType;
    }
 
-   if (error != NoError || error == NeedHostLookup)
+   if (error == NoError || error == NeedHostLookup)
```
jamesu: whoops

Engine/source/platform/platformNet.cpp
pacomont: Talking from the perspective that it's actually working and looking here [connect function (msdn.microsoft.com)](https://msdn.microsoft.com/en-us/library/windows/desktop/ms737625(v=vs.85).aspx) I think with this PR we are doing the things right.

> With a nonblocking socket, the connection attempt cannot be completed immediately. In this case, connect will return SOCKET_ERROR, and WSAGetLastError will return WSAEWOULDBLOCK.

So we should always wait a -1 (SOCKET_ERROR), then we must check if getLastError() returns Net::WouldBlock. Next...

> In this case, there are three possible scenarios:
> Use the select function to determine the completion of the connection request by checking to see if the socket is writeable.
> If the application is using WSAAsyncSelect to indicate interest in connection events, then the application will receive an FD_CONNECT notification indicating that the connect operation is complete (successfully or not).
> If the application is using WSAEventSelect to indicate interest in connection events, then the associated event object will be signaled indicating that the connect operation is complete (successfully or not).

Actually we are doing the first option (using select option), here:

```
bool netSocketWaitForWritable(NetSocket handleFd, S32 timeoutMs)
{  
   fd_set writefds;
   timeval timeout;
   SOCKET socketFd = PlatformNetState::smReservedSocketList.resolve(handleFd);

   FD_ZERO( &writefds );
   FD_SET( socketFd, &writefds );

   timeout.tv_sec = timeoutMs / 1000;
   timeout.tv_usec = ( timeoutMs % 1000 ) * 1000;

   if( select(socketFd + 1, NULL, &writefds, NULL, &timeout) > 0 )
      return true;

   return false;
}
```

jamesu: Keep in mind this code also needs to work for unix/bsd sockets api. Also edge cases can crop up, etc - which is why I was being cautious here. ;)

Taken from the man page on OSX:

	Upon successful completion, a value of 0 is returned.  Otherwise, a value of -1 is returned and the
	     global integer variable errno is set to indicate the error.

	ERRORS
	     The connect() system call will fail if:

	     [EINPROGRESS]      The socket is non-blocking and the connection cannot be completed immediately.  It
	                        is possible to select(2) for completion by selecting the socket for writing.

Which would imply ::connect == 0 or ::connect == -1 and setting errno to EINPROGRESS are the two valid success conditions here.

Thankfully reading through this code, I realize now that it should fall past this block if those conditions are met so this stuff should be ok. Good work 👍

jamesu:
Also if you want to test the ipv6:

* set `$pref::Net::IPV6Enabled` to 1.
* If your using TCPObject and you want it to _listen_ on ipv6, you need to change `NetSocket newTag = Net::openListenPort(port);` to `NetSocket newTag = Net::openListenPort(port, NetAddress::IPV6Address);` (this prob needs a PR so you can specify this from script).
* Connecting TCPObject should just go through Net::openConnectTo which will handle ipv6.
* Make sure your machine is either connected to an ipv6 network or has an address such as `2012::1`, `2012::2` (etc) with subnet prefix length 64.
* Try connecting to ipv6 servers using the notation `IP6:<address>`, `IP6:[<address>]:<port>`, or `IP6:[<address>]:<port>%<interface>`

Engine/source/platform/platformNet.cpp
pacomont: Thanks!! ;)

pacomont:
Actually my configuration are not compatible with IP6... Sorry, but now I can't not make the switch yet. I don't even know if my hardware is compatible...

ipv6

jamesu:
You can make your own private ipv6 network by assigning your own ipv6 addresses in the network configuration settings for one of your ethernet adapters (see my previous post for an example). Should behave the same as if you did it for ipv4 (with the exception of routing to the outside world). For the purposes of testing basic socket functionality & multicast, this is usually sufficient.

pacomont:
I did, but I used an external test program as server [Packet Sender](https://packetsender.com/), which
 lets you start a TCP server using IP6 (and IP4).

As you can see In the image TCPConnect::onConnected(%this) sends a message ("HELLO") which is received by server.  
I have use the localhost on IP6 to make the connection (TCPConnect.connect("IP6:[::1]:8081");). I've used: 0:0:0:0:0:0:0:1 and my local IP address on IP6 also. 
I have made test against the server in IP4 mode also.  

connectip6

*Note: but I can't not understand why the client it's disconnected just intermediately after connection is performed. First is called ::onConnected and then ::onDisconnect. It happens on IP6 and IP4 also. But because my application a use IP4 to connect to other program and the connection is stable it doesn't worry me.

Areloch:
So, we thinking this is good enough as-is? Or did we want to investigate on the ipv6 side of things a little more?

pacomont:
Give me a couple of days and let me check why I have those disconnections. I'm very busy now...

Areloch:
Just figured I'd give this a poke again so it doesn't slip your mind! :)

Whenever you get a chance, just give it a check. I'd like to get this rolled in.

pacomont:
Sorry, I was really busy...

The  only thing left I wanted to check was why those disconnections, just after sending. I have look into Packet Sender and I found "Persistent TCP Connections" was disable. Enabling it, there is no more disconnections... So yes, now this is good enough as-is.

PR #1956, opened 2017-02-25
Enable wind effects by default

Duion:
This enables the forest wind effects by default through initiating it with good default values that work well in most cases.
I always found it annoying to always having to go through every forest item and every one of those wind values and enter some values there so the wind animation starts and in most cases those were very similar values.
It also helps new people since it will be enabled by default with good values, so they do not have to research first what values to enter there, since it is not obvious, that you have to use values in the 0.1 or 0.01 range most of the time.
This way you just need to eventually adjust the values and not set them every time.


dottools:
This is contradictory to your PR #1955 in terms of having options on by default. Pretty sure wind effects are more expensive and troublesome to deal with than foot decals. I suggest these be left disabled and you're just going to have to figure how to propagate a default `ForestItemData` profile to your forest.

Duion:
It is not about performance, but about likeliness you will need it enabled or disabled.
Wind effects you will want enabled in 95-100% of all cases and footsteps you only want to have enabled in maybe 0-5% of all cases.
If you create a game project and the stuff you never need is enabled by default, you end up switching it off by hand hundreds of times and the same for the reverse case.
And if you do not want wind effects, just turn them off or do not place a wind emitter, so you can switch them on and off globally wiht just one setting or per level through the wind emitter.
So your point is not that good.

But I found another much better argument why leave it off, since when newbies do not paint their trees correctly with vertex paint or not at all, the trees will go all wonky, but only if you place a wind emitter.

PR #1957, opened 2017-02-25
Enable playerStep and disable playAmbient by default

Duion:
Set player step enabled by default and play ambient disabled by default.

AllowPlayerStep was set to false one time in hope to fix the tree climb bug, but it did not work or just temporarily, since the bug is back. PlayerStep should be enabled by default, since in 99% of the time you will want it on, since otherwise you get stuck on every little step on your models, even if it is just 1cm, which can be very annoying and frustrating for newbies, since they have to research first what is wrong.

And playAmbient should be false by default as well, since in 99,9% of the time you will not need it on your models and even if you want it, it is problematic, since as soon as you put the model into the scene it starts animating and it is almost impossible to stop it at the root pose again.

### DIFF
--- a/Engine/source/T3D/tsStatic.cpp
+++ b/Engine/source/T3D/tsStatic.cpp
@@ -102,10 +102,10 @@ TSStatic::TSStatic()
    mShapeName        = "";
    mShapeInstance    = NULL;
 
-   mPlayAmbient      = true;
+   mPlayAmbient      = false;
    mAmbientThread    = NULL;
 
-   mAllowPlayerStep = false;
+   mAllowPlayerStep = true;
 
    mConvexList = new Convex;
 


John3:
AllowPlayerStep should be enabled by default. You right, you may encounter this problem even in torque3d game on steam (and players on steam call it "bug" but is not) and some torque3d demos made by newbies.

I not see playAmbient is a problem If you want to disable just hit Rendering->playAmbient. And for reverse to root pose, you go to shape editor and select rootpose.

Duion:
Ambient animations do not have poses at all, it is just a track from frame 0 to whatever and hitting frame 0 is a very slim chance.
And going into the shape editor does not edit the one instance you placed, just the shape in general, so this is not possible.
My intention was just, why have it enabled, if you will only need it in 0.1% of all cases you place a static object?

John3:
Thinking about this, will be good to have playAmbient set to false.

I remember an annoying "bug" from the artist side that screw up all my town, the artist who did the buildings all the models have an little animation bug, all the building move their place with playAmbient set to true, after set to false to all my models and do my town (I never think this was a problem) an old t3d update set to true again and all my town went to hell.

So yes will be better to have it false.

Duion:
Yes, my suggestions come from experience, since some default settings drove me mad so many times I decided to change them for my engine fork.

Areloch:
Was discussing this with az, and this, along with the other default settings PRs, we were contemplating a middleground solution.

When we shift over to assets, the assets can have various properties defined on their by default, and examples of this could be that during the import stage, being able to flag 'hasAmbientAnimation' or the like. You flag that during import(default on, but can be flagged off if you have animations you don't want ambient for that asset), and it's stored to the asset definition as part of the import process. This can be updated later if needbe, but then when spawning your object, it'd get that flag off the asset def and appropriately disable ambient animation(in this example) or whatever else.

In this way, the default presumption of 'if it has animation, you probably want it to play' can be retained, but if you get a case where you have an ambient animation and do not want it to play, this can also be taken care of with the asset definition flag, so you only have to set it once on that specific asset and not worry about it again.

Does that seem like a good compromise approach?

Duion:
Sounds good, but I would still set it by global default to the setting you will use most, does not matter if you can add custom defaults later. The more choices people have to do, the more confusion can arise.
A newbie does not know that he has to turn allowPlayer step on, for being able to step on it, be it in the default way it is now or the asset manager that is to come.

Azaezel:
For defaults, I'd tend to favor further digging on why the whole 'climb up a lightpole like it was a ladder' bug occurred in the first place, then flipping walkables back on, and leaving playAmbient on. From an ergonomics standpoint, that'd be the fewest clicks to handle outliers. And I shoudl think somebody handing you an animated model that they don't want animated *is* an outlier condition.

Duion:
@Azaezel 
You can climb up vertical cylinders like lightpoles and collision meshes for trees, but only on the edge that faces the positive y-axis, in case that helps you.

PR #1958, opened 2017-02-28
Foobarbariansdevbranch

killsap:


Areloch:
Looks like you accidentally committed some stuff from your own project on in there. I'm gunna presume that wasn't intentional. Should this get closed?

PR #1963, opened 2017-03-11
Exiting typo fix

elfprince13:
Fixed a typo, plus lots of white-space fixes.


Areloch:
are the 

ConnectionNotifyEvent*   smConnectionNotify = NULL;
ConnectionAcceptedEvent* smConnectionAccept = NULL;
ConnectionReceiveEvent*  smConnectionReceive = NULL;
PacketReceiveEvent*      smPacketReceive = NULL;

Declarations supposed to have the Net namespace on them, then? I remember we had to do a quick rollback before. I'm not getting any compilation errors or anything, just wanted to check if we were sticking with them being inside the Net namespace, or keeping them global.

Other than that, this PR's looking good.

elfprince13:
They're static members of the Net struct declaration - hence the naming convention of `sm` rather than `g`.

Areloch:
Ok, so the way it is in this PR should be considered correct then. Just wanted to be sure because like said, we'd had to run that quick rollback from before.

elfprince13:
Yeah, this PR should be the correct version.

~Thomas

PR #1970, opened 2017-03-31
squish update (primarily to add additional formats for later usage)

Azaezel:


Areloch:
Seems to be working fine.

PR #1972, opened 2017-04-05
Fix bad initialization on camera

pacomont:
Initialization of y component of rotation. I think It could lead to problems if camera has mDataBlock->cameraCanBank set.


Areloch:
Ahh, good spot!

Issue #1975, opened 2017-04-06
Camera label does not match camera option in WorldEditor


klaus95:
At the right bottom corner in WorldEditor, there is an option to change the camera settings. So, when you change it to 3rd person shooter and close the worldEditor, the moment you reopen the WorldEditor the label of the camera setting is set to 3rd person shooter and the camera is in 1st person shooter.

Steps to replicate the bug:
1. Open WorldEditor
2. Camera setting is on “First Person Shooter"
3. Change camera settings to Third Person Shooter (Label changes and the camera changes)
4. Exit WorldEditor
5. Reopen WorldEditor (Label is set to “Third Person Shooter” while the camera goes back to “First Person Shooter” )

![17693123_1691340340891124_975157578_o](https://cloud.githubusercontent.com/assets/25381649/24737574/cf36ceac-1a5d-11e7-8cb5-c8095fb8e309.png)

PR #1976, opened 2017-04-08
Fix for issue #1951 & #1975

killsap:
#1951 : 
When two instances of Torque3D are running on the same network, one as a
hosted server and the other as a client, the client is unable to connect
to the server. The program now can open a socket. This was obviously a
typo. To find the error I just searched for keywords containing socket,
open, open socket, and opensocket till I found the correct file. The
other error was in how the program processed “non fatal” errors. I just
outputed the error to the main console and devised that the socket
should not be closed for a WSAEWOULDBLOCK error 10035 "It is normal for
WSAEWOULDBLOCK to be reported as the result from calling connect on a
nonblocking SOCK_STREAM socket, since some time must elapse for the
connection to be established."

Read about the WouldBlock error:

https://msdn.microsoft.com/en-us/library/windows/desktop/ms740668(v=vs.85).aspx

Use of dbgSetParameters ( port , password ):

http://docs.garagegames.com/tge/official/content/documentation/Reference/Console%20Functions/TorqueScript_Console_Functions_2.html#dbgSetParameters_.28_port_.2C_password_.29

#1975 : 
The change was made to the function that is executed when entering the world editor. The boolean variable for first person shooter was set to true, since the camera always defaults to first person on entering the world editor. On setting this boolean variable to be true, the correct camera setting will always be displayed on entering the world editor.
[Milestone 6.docx](https://github.com/GarageGames/Torque3D/files/907191/Milestone.6.docx)


dottools:
None of the changes in this PR have anything to do with what is said in the PR description. What's going on here?

killsap:
It was a mistake. :)

PR #1977, opened 2017-04-09
Removes bits of code and includes that are based on old console flags

Areloch:
Removes bits of code and includes that are based on old 360, xbox and PS3 flags that are no longer needed.


Engine/source/console/debugOutputConsumer.h
```diff
@@ -27,8 +27,8 @@
 
 //#define TORQUE_LOCBUILD
 
-#if !defined(TORQUE_DEBUG) && defined(TORQUE_OS_XENON) && !defined(TORQUE_LOCBUILD)
-#define DISABLE_DEBUG_SPEW
+#if !defined(TORQUE_DEBUG) && !defined(TORQUE_LOCBUILD)
+   #define DISABLE_DEBUG_SPEW
```
dottools: Interested to know whether disabling debug spew is necessary anymore since old condition required Xenon to be defined/enabled.

Areloch: I guess it'd largely depend on the circumstance LOCBUILD is flagged. It's a separate flag, so I wasn't sure about trimming it as well if there's some logic behind it. If not though, then yeah could probably just trim that entire bit after all.

Engine/source/ts/tsMeshIntrinsics.cpp
```diff
@@ -59,24 +59,19 @@ MODULE_BEGIN( TSMeshIntrinsics )
...
+            zero_vert_normal_bulk = zero_vert_normal_bulk_SSE;
+         #endif
       }
       else if(Platform::SystemInfo.processor.properties & CPU_PROP_ALTIVEC)
       {
-   #if !defined(TORQUE_OS_XENON) && defined(TORQUE_CPU_PPC)
-         zero_vert_normal_bulk = zero_vert_normal_bulk_gccvec;
-   #endif
+         #if defined(TORQUE_CPU_PPC)
+            zero_vert_normal_bulk = zero_vert_normal_bulk_gccvec;
```
JeffProgrammer: can get rid of PPC as well. Torque doesn't support powerpc architecture / altivec

Azaezel:
Checks out.

Issue #1738, opened 2016-08-11
strip/getTrailingNumber fails with "single character+number"


irei1as:
Script functions stripTrailingNumber("string") and getTrailingNumber("string") don't work correctly if the string is a single character followed by a positive number.

Examples:
Using stripTrailingNumber("t1") returns an empty string instead "t".
Using getTrailingNumber("a22") returns 0 instead of 22.

If the number is negative, both work.
("t-1" and "a-22" return correct values in the previous examples.)

A single space before the number (ex: " 11") fails for strip (empty string) but works for getTrailingNumber.

saxonschmauderer:
Our team seams to think we have fixed the bug. Where can we test our fix within the program? Thanks!

irei1as:
@saxonschmauderer The console functions are after
https://github.com/GarageGames/Torque3D/blob/development/Engine/source/console/consoleFunctions.cpp#L795
But more important is the C++ code that you can find in
https://github.com/GarageGames/Torque3D/blob/development/Engine/source/core/util/str.cpp#L1597

You can just test your fix running the console (pressing ~ key) and typing the examples like:
stripTrailingNumber("t1");
getTrailingNumber("a22");

And see if you get the correct value ("t" and "22").

klaus95:
@irei1as we were able to fix the bug and will commit changes soon.

PR #1983, opened 2017-04-17
Fix for bug #1738

klaus95:
getTrailingNumber("string") and stripTrailingNumber("string") will now work for single letter cases. For example, getTrailingNumber() will return "11" if the input is "a11" or "t11", and stripTrailingNumber() will return "a" if the input is "a11".


Areloch:
Sweet! I'll be checking this out.

PR #1986, opened 2017-04-21
Potential fix for bug #1975

saxonschmauderer:
The change was made to the function that is executed when entering the world editor. The boolean variable for first person shooter was set to true, since the camera always defaults to first person on entering the world editor. On setting this boolean variable to be true, the correct camera setting will always be displayed on entering the world editor. This does not affect other camera labels.


killsap:
Looks good.

PR #1988, opened 2017-04-23
Better bug fix for #1738 (str.cpp)

### LINKED ISSUE #1738: strip/getTrailingNumber fails with "single character+number"

DTFuchs:
Fixes #1738 without redundant checks. Because of the while loop on line 1616 giving the variable p its value, the check against base is redundant.

### DIFF
--- a/Engine/source/core/util/str.cpp
+++ b/Engine/source/core/util/str.cpp
@@ -1620,7 +1620,7 @@ String String::GetTrailingNumber(const char* str, S32& number)
    if ((*p == '-') || (*p == '_'))
       number = -dAtoi(p + 1);
    else
-      number = (isdigit(*p) && (p == base.c_str()) ? dAtoi(p) : dAtoi(++p));
+      number = (isdigit(*p) ? dAtoi(p) : dAtoi(++p));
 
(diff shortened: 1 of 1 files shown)


killsap:
Good code cleanup. Not really a fix. You can read about the fix here: https://github.com/GarageGames/Torque3D/pull/1983

killsap:
"Improved" bug fix. Since the bug was fixed here: https://github.com/GarageGames/Torque3D/pull/1983

Areloch:
Yeah, the #1983 did indeed fix it as killsap said, but you are right, trimming unneeded checks if possible is also good.

I'll be checking this out, thanks!

Issue #1943, opened 2017-02-07
ingame ALT+C to toggle the camera doesn't work


John3:
In game don't work ALT+C to toggle the camera, this worked in v3.8. 

http://pastebin.com/0AUQ72Te

`scripts/server/components/game/camera.cs (134): Unable to find object: 'SGOManager' attempting to call function 'Spawn'
scripts/server/components/game/camera.cs (140): Unable to find object: '' attempting to call function 'getComponent'
scripts/server/components/game/camera.cs (140): Unable to find object: '' attempting to call function 'setupControls'
Set::add: Object "" doesn't exist
SwitchCamera: No client or target camera!
scripts/server/components/game/camera.cs (109): Call to error in switchCamera uses result of void function call.
SwitchControlObject: No client or target controller!
scripts/server/components/game/controlObject.cs (81): Call to error in switchControlObject uses result of void function call.`

killsap:
Why should ALT + C be available in game? It seems like it should only be available in editor.  Also, how did you set up your environment to get this error? Because the ALT + C works in the torque3D source I downloaded and ran in game and editor.

PR #1989, opened 2017-04-24
Added Alt+C to worldEditor.ed.cs

saxonschmauderer:
Moved the Alt+C binding from Templates/Full/game/scripts/server/components/game/camera.cs to worldEditor.ed.cs. Alt+C should only be accessible to the world editor and not the user interface.

### DIFF
--- a/Templates/Full/game/tools/worldEditor/scripts/editors/worldEditor.ed.cs
+++ b/Templates/Full/game/tools/worldEditor/scripts/editors/worldEditor.ed.cs
@@ -20,6 +20,8 @@
 // IN THE SOFTWARE.
 //-----------------------------------------------------------------------------
 
+GlobalActionMap.bind("keyboard", "alt c", "toggleEditorCam");
+
 function WorldEditor::onSelect( %this, %obj )
 {
    EditorTree.addSelection( %obj );


saxonschmauderer:
The previous Alt+C pull request simply removed the functionality for using the Alt+C camera from the game completely. This fix removes the Alt+C camera ONLY from the User interface, and allows normal functionality inside of the World Editor.

Areloch:
Ah, so it moves it so it works in the perview of the editor. Gotcha.

AlexBarys:
To be clear, the previous pull request that I did on this issue did not remove the Alt+C camera toggle functionality from the game completely, as the Alt+c keyboard command was still bound to the correct function to toggle the camera in a different file.  The line(s) that I removed simply removed the re-binding of the alt+c keyboard command to a different and no longer existent function, so that the single correct binding was not overwritten.  However, this does still leave the alt+c camera toggle functionality available when not in the world editor, and just in the game itself.  As noted in the issue #1943, the alt+c camera toggle functionality used to work in the game itself, and it was our intent to restore that functionality. We were unaware that the alt+c camera toggle functionality was only supposed to work with the world editor open, and not in the game itself.  If that is the case, then this pull request is better, but if it is also supposed to work in the game itself, when the world editor is closed, as implied by the issue #1943, then my original pull request accomplishes that where this pull request does not.

I just want to make sure that removing the alt+c functionality from the game and leaving it available only in the world editor is the intended fix for this bug and the intended functionality of the keybind.  If that is the case, then I misunderstood what the intent of the keybind and this pull request is correct.

aaravamudan2014:
I agree with @AlexBarys . There is no point in arguing the validity of these two pulls unless we know why the Alt+C remapping was placed there in the first place.

Azaezel:
If the points to be able to pop out the tools directory and lock players gut of the capacity to hop into fly cam, best to fully segregate out the entire logic chain from the full/basegame templates, since anybody with console access can replicate it. If not, can't say I see the point in forcing the extra click to open the editor then swap, then close it and go on your merry way to see how things would looks from various points of view...

Areloch:
As Az said, I think what'd be best is figure out exactly what the broad plan for the keybind is before deciding the approach for the PR.

The tools directory should indeed be pop-out-able, so having a keybind in the editor dir makes sense at minimum, so it'd be a question of if the regular game should also have the keybind for convenience, and then just dictate the active binds list based on if the editor is open or not.

Issue #1272, opened 2015-04-11
Feature Request: AITurret: IgnoreList - 2 addational Torque Script functions


WaxyChicken:
the ignore list is handled in Torque3D/Engine/source/T3D/turret/aiTurretShape.cpp
the only current functions exposed to TorqueScript are:

```
%AITurretObj.addToIgnoreList(%shapeBaseObj);
%AITurretObj.RemoveFromIgnoreList(%shapeBaseObj);
```

these torquescript functions would add greater control over the turret's ignore list
and would be very useful.

```
torque script function: 

%AITurretObj.ignoreListCount();

return the number of objects in the turrets ignore list.

returns: zero based count of items in the AITurret's ignore list.

example use in torqueScript: 
%ObjectsIgnored = %AITurretObj.ignoreListCount();
```

```
torque script function: %AITurretObj.getIgnoreListObject( int i );
returns: the ignored object ID at index i of the AITurret's ignore list

example use in torqueScript: 
%IgnoredOBJ = %AITurretObj.getIgnoreListObject(2);

example returns the %obj objectID of the item 2 in the ignore list.
```

example use of how both functions can be useful together in TorqueScript:

```
           for(%count = 0; %count < %AITurretObj.ignoreListCount(); %count++) 
           {
               %AITurretObj.getIgnoreListObject(%count).health += 10;
           }
```

eightyeight:
See also #1271

eightyeight:
We'd be happy to accept pull-requests with these changes! :)

PR #1991, opened 2017-04-25
Ai turret ignore list functions

AlexBarys:
Implements feature(s) requested in issue #1272 
The first feature is exposing a function to Torquescript that allows retrieving the count of the total number of objects on a turret's ai ignore list, ignoreListCount().

The second feature is exposing a function to Torquescript that allows retrieving a reference to an object on an AI turret's ignore list based off of the index in the list of that object, rather than the object's ID (so the object's ID does not have to be known beforehand).  The feature request issue #1272 requests for this feature to return the object's ID but then shows an example modifying the object directly (increasing the health of the object by 10), so to fulfill the requested functionality based on the example, this feature returns a reference to the object itself so that properties can be called and modified on the object directly, as shown in the example on the feature request.

### DIFF
--- a/Engine/source/T3D/turret/aiTurretShape.cpp
+++ b/Engine/source/T3D/turret/aiTurretShape.cpp
@@ -564,6 +564,16 @@ void AITurretShape::removeFromIgnoreList(ShapeBase* obj)
    mIgnoreObjects.removeObject(obj);
 }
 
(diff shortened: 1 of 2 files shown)


dottools:
Makes sense. Looks good, thanks again.

Areloch:
Thanks for this!

Issue #1498, opened 2016-01-12
Decal Properties Wrongly Name In Script


SteveYorkshire:
LOD for decals can be generated but the properties are wrongly named in the script file.

https://github.com/GarageGames/Torque3D/blob/master/Templates/Full/game/art/decals/managedDecalData.cs#L42-L43
and
https://github.com/GarageGames/Torque3D/blob/master/Templates/Full/game/art/decals/managedDecalData.cs#L54-L55

Decals with properties are not referenced in the source:
   screenStartRadius
   screenEndRadius

Those  should be the properties referenced in the source which are:
   fadeStartPixelSize
   fadeEndPixelSize

https://github.com/GarageGames/Torque3D/blob/master/Engine/source/T3D/decal/decalData.cpp#L89-L90

My apologies for not putting in a pull request, I'm too autistic to be a normie but not autistic enough to use github without pulling my hair out ;)

PR #1992, opened 2017-04-25
Fixed old names of decal properties

AlexBarys:
Renamed the screenStartRadius and screenEndRadius properties in the bulletHoleDecal datablock and ScorchRXDecal datablock to fadeStartPixelSize and fadeEndPixelSize properties, respectively, to match names of the properties as defined in the decalData.cpp file, as noted in issue #1498
Among other possible fixes, this makes the size of the bullet hole decal's actually random in size now, instead of always being the exact same size.


Areloch:
Looks good, thanks!

Issue #1501, opened 2016-01-14
dragCoefficient misspelt 6 times over 2 scripts


SteveYorkshire:
dragCoeffiecient??? One "e" too many. Should be "dragCoefficient"

Offending entries:
https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/game/art/datablocks/weapons/rocketfx.cs#L31
https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/game/art/datablocks/weapons/rocketfx.cs#L31
https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/game/art/datablocks/weapons/grenadefx.cs#L301
https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/game/art/datablocks/weapons/grenadefx.cs#L465
https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/game/art/datablocks/weapons/grenadefx.cs#L633
https://github.com/GarageGames/Torque3D/blob/development/Templates/Full/game/art/datablocks/weapons/grenadefx.cs#L848

Happens 6 times but only 1 has a setting that is not zero. Not handy if you're gonna copy and paste example scripts though. ;)

ztb1020:
I made fixes to all of these.

PR #1993, opened 2017-04-25
Drag coefficient misspelled

AlexBarys:
This addresses issue #1501 
The property variable "dragCoefficient" was misspelled in various datablocks in the grenadeFX.cs and rocketFX.cs files, both in the Templates directory locations and the Modules directory locations.  It was misspelled as "dragCoeffiecient", as noted in the issue ticket.  Some but not all of those locations were spelled correct, but the property was misspelled in numerous other locations in those two files as well.  This corrects the spelling of those properties so that users trying to reference those properties in other parts of the Torquescript will be able to reference them by the correct name.  
For instance, if one had a list of datablocks that included some where the property was spelled correctly and some where it was spelled incorrectly, and wanted to iterate through that list, changing the "dragCoefficient"property for each, they would get an error as some of the datablocks would not have the "dragCoefficient" property, as it was misspelled in them.
This also makes it more intuitive for users to reference that property in other parts of Torquescript even when referencing it for a single datablock as they now do not need to know that the property is misspelled, and how it is misspelled.

### DIFF
--- a/Templates/Full/game/art/datablocks/weapons/grenadefx.cs
+++ b/Templates/Full/game/art/datablocks/weapons/grenadefx.cs
@@ -55,7 +55,7 @@ datablock LightDescription(GrenadeLauncherLightDesc)
 datablock ParticleData(GrenadeDebrisFireParticle)
 {
    textureName = "art/particles/impact";
-   dragCoeffiecient = 0;
+   dragCoefficient = 0;
    gravityCoefficient = -1.00366;
    inheritedVelFactor = 0.0;
    constantAcceleration = 0.0;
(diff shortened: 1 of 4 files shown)


dottools:
Oops. Overall looks good. I would suggest [squashing it down to one commit](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit) though. You can do that without a new PR as any changes in your source repository branch will be reflected in this PR.

Areloch:
As @dottools said, looks good, just hoping you can squash the history to make it a cleaner PR first.

Thanks!

AlexBarys:
Sorry for the delay, it should now be squashed into one single commit in my branch.
Thanks for the feedback.

PR #2000, opened 2017-04-30
Fix for bug #1975

saxonschmauderer:
The change was made to the function that is executed when entering the world editor. The boolean variable for first person shooter was set to true, since the camera always defaults to first person on entering the world editor. On setting this boolean variable to be true, the correct camera setting will always be displayed on entering the world editor. This does not affect other camera labels.


saxonschmauderer:
@Areloch here is the corrected pull request without the two additional Alt+C file changes added. Still trying to figure out Github. Thanks!

Areloch:
Thanks, and no worries about taking a bit to get to speed with github. Takes a little while for everyone :)

I'll be checking this out soon.

PR #2001, opened 2017-05-02
Why not apply Camera FX on vehicle driver position?

pacomont:


Areloch:
Good question! I'll check this out, thanks.

PR #2002, opened 2017-05-07
VolumetricFog memory leak fix

OTHGMars:
Fixes a memory leak on clients and prevents a crash from deleting a RTT target while it's still registered with the GFX device.


Areloch:
Ah, nice!

PR #2007, opened 2017-05-12
BugFix for GFXDrawUtil::drawTextN()

OTHGMars:
This PR fixes a bug that was causing GFXDrawUtil::drawTextN() to draw one character more than requested if in_string is longer than n. This bug shows up in the GuiMessageVectorCtrl when lines are split. The first character of the second part of the string is duplicated on the line above. You can trace through from here: https://github.com/GarageGames/Torque3D/blob/development/Engine/source/gui/game/guiMessageVectorCtrl.cpp#L717 to see it happen, if you have a control with split lines.
Specifically,
GFX->getDrawUtil()->drawTextN(mProfile->mFont, start, "0123456789", 5, mProfile->mFontColors, mMaxColorIndex, 0.0f);
will render 6 characters instead of 5 as requested.


dottools:
Nice catch. Thanks for the PR.

Areloch:
Not noticing any issues with this. Thanks!

Issue #2005, opened 2017-05-09
dumpConsoleClasses() crashes torque


andijh92:
dumpConsoleClasses() crashes torque.

i was using: 
ubuntu 16.04.
torque 3.10 (last commit c460868814f1b41c954fdd963ea9a7f60c017f18 )
self compiled binaries and default game template via $ make install

Tail of the output: https://pastebin.com/nuehK2in

dumpConsoleFunctions(), dumpEngineDocs("foo") and playing a little bit the default game template worked.

Areloch:
I've figured out where it crashes, I've mostly been looking through for the 'why' so it can get fixed.

Areloch:
Dunno if you spotted the PR, but if you could eyeball #2008 and see if that fixes the crash on your end, I can get it rolled in.

andijh92:
Sry, for the delay! Now it's tested.

Areloch:
Fixed

PR #2008, opened 2017-05-14
Console Classes dump allocation fix

Areloch:
Console Classes dump fix. It was running aground on having a case where there was a empty string value for the 'usage' field in the class header. This would break the allocation because we're allocating nothing. This rejects it if the usage field is an empty string.

This is in reference to issue #2005.

### DIFF
--- a/Engine/source/console/consoleDoc.cpp
+++ b/Engine/source/console/consoleDoc.cpp
@@ -85,7 +85,7 @@ void printClassHeader(const char* usage, const char * className, const char * su
       Con::printf("///       information was available for this class.");
    }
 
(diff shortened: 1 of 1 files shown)


dottools:
Can't compare a char pointer to an empty char array constant. Have to use `strlen()` instead:
```C++
if((usage != NULL) && strlen(usage))
```

Areloch:
It did work, but checking the length is probably smarter. Good idea, I'll fix that.

andijh92:
`if((usage != NULL) && strlen(usage)) `worked for me.
`if( usage != NULL && usage != "")` caused a crash.

PR #2017, opened 2017-05-28
Removes Direct3D9 functionality.

Areloch:
Removes the aging D3D9 support in favor of better functionality support for modern OGL and D3D11.


dottools:
RIP.

Areloch:
Yep. It's served incredibly well, but we gotta strip it out so we can better support the new stuff.

PR #2018, opened 2017-06-02
ForceFOV is not working on GuiTSControl.

pacomont:
The definition of frustum is already made. Changing mLastCameraQuery.fov on _internalRender is ignored. It has to be changed before frustum is defined.

Note: With this PR ForceFOV will not work on Render Stereo modes. I don't know it this is important...


Areloch:
Cool.

And for the render stereo bit, it probably is important, I'll test it with it enabled and see if it breaks anything though. Thanks for the note on that.

PR #2025, opened 2017-06-24
Fix load texture warnmat

John3:


rextimmy:
Aah nice pickup! The path to this warning texture probably should be set as a global var in script land rather than hard coded in everywhere.

John3:
Thanks timmy, I will added like gloval var.

the same have with the editor shapes.

https://github.com/GarageGames/Torque3D/pull/2028/commits/ed8b87d51657c8e5d2d321b7bea7402dd2becd64

PR #2028, opened 2017-06-24
the remaining shapes need go inside the tools folder

John3:


Areloch:
I'd say the only exception to the move would be the noShape. That'd be a good, general fallback shape to display if a 3d model is missing(similar to how we have the warn mat). Everything else moving to tools makes sense to me though!

John3:
new PR https://github.com/GarageGames/Torque3D/pull/2042

PR #2030, opened 2017-06-29
VS2017 compiler workaround

Areloch:
Works around the regressive behavior in the VSC++ compiler for VS2017 where the variadic templates would not compile correctly. Issue should be patched in future versions, so for now it'll target the 1910 version specifically.

### DIFF
--- a/Engine/source/console/engineFunctions.h
+++ b/Engine/source/console/engineFunctions.h
@@ -108,7 +108,17 @@ private:
       std::tie(std::get<I + (sizeof...(ArgTs) - sizeof...(TailTs))>(args)...) = defaultArgs;
    }
    
+#ifdef _MSC_VER == 1910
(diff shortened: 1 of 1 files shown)


dottools:
Ah Visual Studio. Still the only compiler that requires dodgy workarounds. ;)

#MakeTemplatesGreatAgain

rextimmy:
👍  for leaving my DodgyVCHelper name in place :). Just a note about this, MS claim this is to be fixed with VS 15.3 release, so far testing 15.3 preview release the issue is not fixed, if the public release of 15.3 the issue is not fixed, the target will version will need to include 1911

Areloch:
Haha. Yeah, what's annoying is MS is aware of this, flagged the bug report as fixed in a future version and apparently it was slated for the .3 update, but it's not in the preview. So....yeaaah.

PR #2036, opened 2017-07-07
Fix light rendering in reflections.

Areloch:
Includes a fix to get lights to render more correctly in the reflection pass. Also includes a helper function to force a render from a passed in transform and frustum.


Areloch:
Win side works fine. Just gotta test it on the linux/mac side to be sure.

PR #2042, opened 2017-07-09
move worldEditor shapes into tools folder

John3:
without noShape


John3:
Closed, I need to found other way to clean a little more the modules, to have just the necessary.

PR #2043, opened 2017-07-11
cleaning the modules

John3:
cleaning the FPS and spectator modules


John3:
@Areloch any feedback?

PR #2047, opened 2017-07-19
extra option for terrain block to disable the update of the basetexture

Duion:
This adds another checkbox in the terrain datablock in the editor, where you can disable the auto generation of the basetexture, which allows you to use a custom basetexture for the whole terrain. Terrain generation software often puts out a precalculated basetexture and with this option you can use it by just renaming it into level.basetex.png or whatever and the engine will use it and not overwrite it.


Duion:
I just noticed this does not prevent the generation on the fly if you paint the terrain in the editor, it just prevents it from saving and overwriting your eventually custom made yourlevel_basetex.dds
So this method could be extended to prevent that as well, but I did not need it that much so far, I just needed it to stop overwriting the file.

PR #2048, opened 2017-07-19
extra option for terrain block to disable the update of the basetextu…

Duion:
…re 2

Little change that was missing in the first PR since Github web interface does not seem allow editing multiple files for a PR and I was lazy.


Areloch:
So just to be sure, this is the one we should use, not the other one, right? If so, I can close the other.

Duion:
You need both PR, I did it with the web interface and it just allows you to change one file at a time.

Issue #1704, opened 2016-07-27
Hard edges on terrain where terrain textures appearance changes


Duion:
Additional to the terrain textures being wrong I noticed that on certain tiles in the map, the terrain rendering suddenly is different. There is a straight line without blending and inside that tile the textures look more like they looked originally while around they are like the 3.9 high contrasty wrong textures.
So it is hard to adjust them if there is those tiles where everything is different.
Here are some example screenshots:
http://duion.com/files/temp/screenshot_TG_DesertRuins_day001-00003.png
http://duion.com/files/temp/screenshot_TG_BuildingSite_day001-00004.png
http://duion.com/files/temp/screenshot_TG_BuildingSite_day001-00007.png

Duion:
News on this issue, I found out that this bug only happens for one specific texture layer, this is why it is so hard to find, since all other terrain layers may blend fine, just one of them in one or more tiles blends wrong and it only blends wrong on itself, means it blends well to all other layers, but not to itself in those bugged tiles.
For example you can paint over the hard edge with the blend issue with another layer and everything blends fine again.
Another thing what you can do is, you can change the order of the terrain layers in the list and suddenly another layer has the blending issues. You may also see whole layers changing in intensity based on where they are in the ordered terrain textures list, so some textures may look fine, others ugly.

Duion:
More infos, it only happens with normal map shader activated and on the high contrast side next to the edge you have those normal map glitch lines.

Duion:
False alert, seam is there without macro texture as well, just not visible that much but the issue consists without macro texture.

Bloodknight:
Just to add to this, i was having the same issue when panning forwards and backwards, it appeared only when a macro texture was in use at the time, removing the macro texture removed the issue at least for me.

Duion:
Azaezels pull request fixes it, what remains is the normal map artifacts where the edge was, but the diffuse blending has no more issues.

PR #2049, opened 2017-07-19
Fixes blending issues while respecting #1704

lukaspj:
The corrections to the blendTotal were reverted due to issue #1704 . This will revert that revert, while also making the terrain render in a single pass (at least for up to ~20 textures per terrain cell, which should be reasonable).

In order to allow rendering the terrain in a single pass, I had to move some calculations from vertex to pixel shader, so a benchmark on how this affects performance should definetly be considered.

In theory, it should be possible to solve this issue with the multi-pass rendering technique utilized currently, but I haven't been able to make it work.

### DIFF
--- a/Engine/source/lighting/advanced/hlsl/advancedLightingFeaturesHLSL.cpp
+++ b/Engine/source/lighting/advanced/hlsl/advancedLightingFeaturesHLSL.cpp
@@ -30,6 +30,7 @@
 #include "renderInstance/renderDeferredMgr.h"
 #include "materials/processedMaterial.h"
 #include "materials/materialFeatureTypes.h"
+#include "terrain/terrFeatureTypes.h"
 
 
 void DeferredRTLightingFeatHLSL::processPixMacros( Vector<GFXShaderMacro> &macros, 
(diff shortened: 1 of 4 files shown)


Areloch:
Sorry it took a bit to get this tested properly.

Looks like something's going oddball in the math with this. Using the stock textures, the blend breaks down when you get several materials in a small area(i've got 6 in a relatively small area on the terrain)

Screens:
(http://ghc-games.com/public/nuBlending01.png)
(http://ghc-games.com/public/nuBlending02.png)
(http://ghc-games.com/public/nuBlending03.png)
(http://ghc-games.com/public/nuBlending04.png)
(http://ghc-games.com/public/nuBlending05.png)

So we're getting a total burn out on the grass texture(detail goes full black, though the base texture is fine), which is also the first material, if that matters. The others look like they have their seams become jagged instead of smooth, we see hard seaming between blocks, and the detail textures look to be wrongly assigned(likely related to the first point. If our detail textures are being mis-indexed, it's possible that our first materials' detail texture is being pointed to an invalid point and doing black).

Areloch:
Double checked. Behavior persists in Release as well. Tried the GL side. Fatal crash in debug, no actual error message though, just crashes in the GFXGLShader initHandles(). Looks like line 662, where it loops through the ConstantDescriptions. In GL Release, the blocks stop rendering as you get close to them, as if some sort of bizzare inverse LOD was happening.

lukaspj:
I believe az mentioned that it'd be possible to get the constNum of textures soon? Maybe we should wait till that is possible and change the code so it doesn't rely on a specific order of textures.

I will try and see if I can figure out what goes on with openGL. 

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

Areloch:
Ah, yeah, I remember you bringing up the constNum bit before. We'll definitely want to get that pinned down.

PR #2055, opened 2017-07-24
Reworks the $Core::commonShaderPath variable usage

Areloch:
Reworks the $Core::commonShaderPath variable usage to be a static variable in ShaderGen for efficiency(only one getVariable when shadergen is initialized), as well as implements the ability to set a default value, and ensures that it tries to set a path even if the pref variable is missing which is important for ported projects from older builds.

### DIFF
--- a/Engine/source/shaderGen/shaderGen.cpp
+++ b/Engine/source/shaderGen/shaderGen.cpp
@@ -49,7 +49,7 @@ MODULE_BEGIN( ShaderGen )
 
 MODULE_END;
 
-String ShaderGen::smCommonShaderPath(Con::getVariable("$Core::CommonShaderPath", "shaders/common"));
+String ShaderGen::smCommonShaderPath("shaders/common");
 
 ShaderGen::ShaderGen()
 {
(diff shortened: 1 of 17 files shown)


Engine/source/shaderGen/shaderGen.cpp
```diff
@@ -49,6 +49,7 @@ MODULE_BEGIN( ShaderGen )
 
 MODULE_END;
 
+String ShaderGen::smCommonShaderPath(Con::getVariable("$Core::CommonShaderPath", "shaders/common"));
```
dottools: This will always evaluate to "shaders/common" as console subsystem won't be initialized until way later, unless it segfaults. Keep in mind when you do object initialization like this at the source file scope it'll be executed before `int main(size_t argc, char *argv[])` is executed. Shouldn't call engine functions until their respective subsystems are initialized.

Areloch: It actually does set it when shaderGen itself is initialized, here: https://github.com/GarageGames/Torque3D/pull/2055/files#diff-2d6f2326413f62282d66e1c77566a3c7R129

So it works fine, but you mentioning it does highlight how silly having the getVariable in that initial definition of the var is. It should just be 
`String ShaderGen::smCommonShaderPath("shaders/common");`

Got too eager with the copy/paste, there ;)

I'll get that updated.

PR #2057, opened 2017-07-27
Adds missing GFX Texture format strings

OTHGMars:
This commit eliminates 5 warnings output from VALIDATE_LOOKUPTABLE here: https://github.com/GarageGames/Torque3D/blob/development/Engine/source/gfx/gfxStringEnumTranslate.cpp#L167
And prevents a fatal error here: https://github.com/GarageGames/Torque3D/blob/development/Engine/source/gfx/gfxTextureManager.cpp#L1371
that occurs when called from GuiOffscreenCanvas::_setupTargets() for an OpenVROverlay because GFXFormatR8G8B8A8_LINEAR_FORCE  is used.


rextimmy:
Nice a few less warnings in the console, tested ok here.

PR #2059, opened 2017-07-30
#define AUTOPAINT_UNDO 

Bloodknight:
This edit makes it easy to enable/disable the terrain autopaint undo.

When using large terrains 1024+ the undo code takes an absolute age to complete, while the undo functionality can be disabled, it is spread throughout the code.


Areloch:
Ah, nice :D

PR #2060, opened 2017-07-30
SQLite Integration 

Bloodknight:
I have used the updated version of Chris Calef work to add this.

I have added the cmake file so that you can choose to have it or not


Areloch:
This seems to work as far as I can see.

I did note that you've got the 'MegaMotion' name in the one pref var, here:

https://github.com/Bloodknight/Torque3D/blob/44d6bf79b9fc9715ffdb1030359b7a18bdc2a6e1/Engine/source/sqlite/SQLiteObject.cc#L226

This probably should be changed to something more general? 

$pref::SQLite::SQLVerbose, perhaps?

PR #2062, opened 2017-08-08
OPCODE lib - rewrite swap, add xorswap

Bloodknight:


Areloch:
Looks good. Only nitpicks I would have are the extra indent throwing the formatting off, and just wondering if we should have the int be their custom udword typedef just for consistency's sake?

Other than that, didn't note any issues.

rextimmy:
Yes it should be udword. udword is an unsigned int so possible loss of data there.

Bloodknight:
@rextimmy done

@Areloch indent looks fine in Visual Studio, additionally, have you seen the indenting in this library?  
If there was ever a set of source files more in need of beautification formatting I've not seen them :P

Areloch:
@Bloodknight  Haha, fair enough. If it's just git choking on the formatting a bit, I'm not going to worry too much because the libs are definitely not consistently formatted. They should probably be spruced, but we'll leave that for a separate PR ;)

PR #2076, opened 2017-09-02
allows different waypoints

John3:
I wanted to share this for some time, not sure if this is the best way to do it, but before it falls into the void :) this will allows to have at the same time different waypoint/markers, add more "WayPointMarker"....

Well you could always add from script, but this PR allows to add it from the World Editor.

See an example, at the right the green old shape. Now you can used for enemies, allies, objects.... I can do other PR with the markers I use, it have different shapes and colors.

markers

markers2

### DIFF
--- a/Templates/BaseGame/game/tools/worldEditor/gui/objectBuilderGui.ed.gui
+++ b/Templates/BaseGame/game/tools/worldEditor/gui/objectBuilderGui.ed.gui
@@ -1024,20 +1024,21 @@ function StaticShapeData::create(%data)
 
 function MissionMarkerData::create(%block)
 {
-   switch$(%block)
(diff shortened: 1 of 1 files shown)


Bloodknight:
can you add the script for the samples you used? This would add a launching point for others to see how it works and add extras as needed.

John3:
sorry man I don't remember, this pr is more than 1 year old :grimacing: but thank you for you interest... For the scripts, I need to search in my old prototype demo. 

As said in the description, the main idea is use the waypoint with different shapes and colors, its depends how you use the waypoints, for example to differentiate visually in the world editor the spawn group for enemies and friends... it can help for a strategy game or for items in single player game

PR #2080, opened 2017-09-12
Fixes smCommonShaderPath init order

### LINKED ISSUE #2079: Shader include path bug in BaseGame template.
Today I tried to compile T3D with the BaseGame template, after i hit the “Launch World Editor” button, some errors occurred. Some  dialogs popped with the info "Failed to open the include 'shaders/common/torque.hlsl'". i think this is a path bug, may be the developers forgot to update the include file path in some shader files, casue the "torque.hlsl" is no longer in the "shaders/common" path but in the "shaders/" path in the BaseGame template.

Here is the post in the forum with  screenshots. Hope that  helps .

Areloch:
Moves the initialization of the shader common path const var so it is set before we do an initial setup of some shadergen fields.

Resolves #2079

### DIFF
--- a/Engine/source/shaderGen/shaderGen.cpp
+++ b/Engine/source/shaderGen/shaderGen.cpp
@@ -96,6 +96,8 @@ void ShaderGen::initShaderGen()
    if (!mInitDelegates[adapterType])
       return;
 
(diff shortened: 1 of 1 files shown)


rextimmy:
Tested ok here

PR #2081, opened 2017-09-13
Standardizes gui controls that load bitmaps to use GFXTexturePersiste…

OTHGMars:
…ntSRGBProfile and removes GFXGuiCursorProfile and GFXDefaultGUIProfile


dottools:
Pretty sure `GFXTextureProfile::NoPadding` is important for skinned controls. And `Persistent` part of `GFXTexturePersistentSRGBProfile` means it uses `GFXTextureProfile::KeepBitmap` which means it keeps the original bitmap loaded in memory even when not required, majority of graphics operations don't need the original bitmap.

Are you sure you fully tested and understand these changes?

PR #2095, opened 2017-10-15
Enables setting the forward vector of a Scene Object

Areloch:
Enables setting the forward vector of a Scene Object

### DIFF
--- a/Engine/source/scene/sceneObject.cpp
+++ b/Engine/source/scene/sceneObject.cpp
@@ -433,6 +433,64 @@ void SceneObject::setScale( const VectorF &scale )
    setMaskBits( ScaleMask );
 }
 
+void SceneObject::setForwardVector(VectorF newForward, VectorF upVector)
+{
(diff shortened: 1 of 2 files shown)


Engine/source/scene/sceneObject.cpp
```diff
@@ -1458,3 +1516,10 @@ DefineEngineMethod( SceneObject, isGlobalBounds, bool, (),,
 {
    return object->isGlobalBounds();
 }
+
+DefineConsoleMethod(SceneObject, setForwardVector, void, (VectorF newForward, VectorF upVector), (VectorF(0, 0, 0), VectorF(0, 0, 1)),
+   "Get the number of static fields on the object.\n"
+   "@return The number of static fields defined on the object.")
```
dottools: Comment needs updating. :)

Areloch:
@dottools Thanks for the spot on that, corrected

PR #2099, opened 2017-10-15
Removes Entity/Component stuff from being behind an experimental flag.

Areloch:
Removes Entity/Component stuff from being behind an experimental flag.


OTHGMars:
This commit fails to compile here if the TORQUE_EXTENDED_MOVE flag is turned on, or if the TORQUE_DISABLE_MEMORY_MANAGER flag is turned off.

Areloch:
Thanks for the spot. Removed the EXTENDED_MOVE block because we'll be doing a VR controller component in the future anyways.

For the rapidjson stuff, it looks broken in stock with the disable_memory_manager flag flipped off, so I'll consider that a separate issue to this that needs fixing.

PR #2103, opened 2017-10-18
Adds functionality to automate convex proxy objects

Areloch:
Adds functionality to automate convex proxy objects for appropriate types. Instead of having to manually convert a zone or portal or occlusion volume to convex to edit it's shape, then back again, simply entering the convex editor will automatically convert them to proxy convex shapes, with a special material to denote them.

When the convex shape editor is exited, it converts them back to their original type, with the changes made to the proxy convex applied to them.


dottools:
Very nice.

PR #2118, opened 2017-11-11
Refactors the Popup menus and GuiMenuBars

Areloch:
Refactors the Popup menus and GuiMenuBars to remove unneeded duplication and platform-specific/deprecated code.


blackwc:
Works for me.

PR #2119, opened 2017-11-12
SFX Player fixes #1544

### LINKED ISSUE #1544: SFX playlist / Bitstream buffer
http://forums.torque3d.org/viewtopic.php?f=10&t=548
SFX Playlist currently writes an absurd amount of data into the bitstream. Decreasing the the max slot count or increasing the buffer size are two possible fixes that have been brought up.

A secondary topic is the large amount of nothing the playlist and some other datablocks write. Expanding pack/unpack functions to detect unused fields/slots could decrease loading times when connecting to remote servers with many datablocks.

blackwc:


chaigler:
@blackwc How does this compare to PR #1509?

PR #2122, opened 2017-11-13
linux SDL canvas offset bug

### LINKED ISSUE #2117: Linux SDL canvas offset problem
If the window resolution is larger than the system resolution, it will resize the window to fit but there is an offset equal to the window border + whatever system panels are taking up the desktop area. The offset appears to only affect the drawing because the hit area for the buttons are where they should be.

The offset is fixed if you resize the window however if you change the window resolution it knocks the offset back.

Attached is a screenshot from Uebergame which shows the offset.

screenshot from 2017-11-03 15-24-31

blackwc:
fixes #2117 

Where the canvas was at offset when window resolution was greater than system resolution.


blackwc:
As has been pointed out, the padding won't account for a menu bar that uses a font much larger than the standard used in menu for GUI editor and world editor tools. I could call for menu height but no menu exists yet. I suggest we merge the padding hack until I can figure out exactly what the heck is going on with the canvas when a menu is added.

blackwc:
The hack is no longer necessary.

blackwc:
New problem resulting from this. Now the offset is positive and only occurs when you resize manually. This is the polar opposite problem of the problem I was attempting to solve in the first place.

blackwc:
I don't know why the magic number 10 is necessary for it to play nice in gnome and fluxbox but it is. The magic number hack makes for smoother experience on linux than without it at least on those two managers.

PR #2132, opened 2017-12-08
Improve the consistency of the EngineAPI

lukaspj:
This will make engineAPI a bit more consistent with which functions it exposes to as C-style functions and the name of those functions.


lukaspj:
This will be contained in #2136. Easier to manage the changes in a single PR.

PR #2144, opened 2017-12-22
Allow correct display of widescreen resolutions (16:9 and 16:10 ratio) on a 4:3 ratio display

RichardsGameStudio:
Using DXGI_MODE_SCALING_UNSPECIFIED will result in a scaling of widescreen resolutions with a width:height ratio of 16:9 or 16:10 on 4:3 ratio displays, changing the resolution to the next available 4:3 resolution, causing distortion of the view. Changing DXGI_MODE_SCALING_UNSPECIFIED to DXGI_MODE_SCALING_CENTERED will prevent this behavior, resulting in a correct display of these widescreen resolutions.


RichardsGameStudio:
Thanks to @Duion for pointing me to this flaw.

Duion:
Cannot remember pointing you to that.

PR #2181, opened 2018-01-28
fix menuButtonHover/Pressed

John3:
SFXProfile should be datablock not singleton


Azaezel:
Depends on if it's for a clientside or server sound. Server-transmitted you'd be right. Menu sounds aren't.

John3:
good to know :) I don't remember read this in the documentation. Thanks

PR #2182, opened 2018-01-28
Asset browser initial

Areloch:
The initial implementation pass for the Asset Browser, and relevant supporting changes to assets, modules, some UI controls, entities and components, and the editors.


Areloch:
There's definitely some issues and bugs still lurking in here I haven't 100% nailed down, but it's broadly stable enough to be rolled in, considered as 'experimental' until the last little bits and bobs get smoothed out going forward.

PR #2186, opened 2018-02-03
Physx 3.4

rextimmy:
PhysX 3.4 support


OTHGMars:
I tested this PR with today's cmake change and the current PhysX 3.4 repo in VS2015 x64. The install projects copied all needed dlls. The engine built and executed without any problems.

Issue #2127, opened 2017-12-06
game don't run with script compile .DSO


John3:
Hi,

I was testing a prototype game and I found I can't run the game if all my *.cs was compiled to *.dso. I need to delete all the dso and leave the scripts to run the game.

I'm using the new template basegame and FPSGameplay module. I did a quick test in the old Full template and don't work.

I'm using last dev https://github.com/GarageGames/Torque3D/commit/74b7e4cd89f4a0fee65d3f831c35df349d16f8c3

Anyone now something about this issue?

Areloch:
Hm, news to me. Definitely needs to be checked out. Could be stuff with the torquescript updates not being agreeable to DSOs due to a bug?

JeffProgrammer:
i'll look into it. might not be for a couple days due to finals w/ school though. thanks john for testing!

JeffProgrammer:
@John3 we believe that it is a 64bit issue. 32bit seems to work fine. I will keep you posted once we have a solution! :)

JeffProgrammer:
All,
I am still trying to come up with a solution for this. It is more complicated than I originally thought.

Azaezel:
elaborate?

JeffProgrammer:
It's a 64bit issue. The code/wordsize assumes 32bit in a few places within the interpreter.  I need to review it again, as I forget exactly where it is. Once I look at it I will post more info. Sorry!

chaigler:
@John3 - Close this one? Looks like it's been fixed?

John3:
Hey @chaigler It's seems that way. thank you.

PR #2187, opened 2018-02-03
Fix Torque3D 64bit DSOs.

JeffProgrammer:
Fix from Torque2D for 64bit string table entries.


JeffProgrammer:
This should fix #2127

John3:
it works :) thx Hutch. Only tested in win, I will test in linux during the day

John3:
Tested on Linux 64 bit :) 👍

PR #2188, opened 2018-02-03
gcc/clang build fix

rextimmy:
Fixes a compile problem with gcc & clang


John3:
thx tim tested on linux, fixed the compilation issue. 👍

PR #2193, opened 2018-02-05
bullet 2.87 update

John3:
update to bullet 2.87


Areloch:
Hm, not compiling for me. It's complaining about "m_multiSapParentProxy" being undeclared in btMultiSapBroadphase.h and .cpp

Are we just not supposed to include those files? Or are we doing something funny elsewhere.

The errors relating to that var are the only ones I got though, so if we get that figured, the rest is likely fine.

John3:
I tested again cleaning everything, it work for me.

PR #2219, opened 2018-03-08
Security Fixes: String Buffers

CouleeApps:
Lots of places in the engine were using strcpy and strcat directly, not respecting the bounds of the buffers they were using. This PR does the following:

 - Cleans up the usages of specifically strcpy/strcat/dStrcpy/dStrcat
 - Redefines dStrcpy/dStrcat to take a buffer size parameter (breaking change)
 - Updates dStrcpy/dStrcat to use strlcpy and strlcat which respect buffer size and warn about overflow
- Fixes multiple buffer overflows that allowed arbitrary code execution or crashed the engine
- Removes legacy macOS code that would cause buffer overflows if it was ever used (it wasn't)


dottools:
These all do look good, however I advise against duplicating buffer length code but instead use a length variable to avoid copy&paste errors, or worst code; somebody after you revises the buffer lengths on allocation but didn't notice length code being used elsewhere and thus creates a new bug.

For example, instead of this:
```cpp
      newDesc = new char[dStrlen(desc) + 1];
      dStrcpy(newDesc, desc, dStrlen(desc) + 1);
```
Do this:
```cpp
      size_t size = dStrlen(desc) + 1;
      newDesc = new char[size];
      dStrcpy(newDesc, desc, size);
```

Makes the intention straight forward to other developers. If ever necessary, instead of multiple `dStrlen(desc) + 1`, only `size_t size = dStrlen(desc) + 1;` has to be changed for expected results instead of surprise bugs. :)

CouleeApps:
That's a good point. I'll fix that soon and add a commit to the pr.

PR #2225, opened 2018-03-18
stop linux/mac  undefined MSVC versioncheck spam.

Azaezel:


CouleeApps:
Here's another: https://github.com/GarageGames/Torque3D/blob/development/Engine/source/console/consoleTypes.h#L51

PR #2136, opened 2017-12-11
WIP: CInterface Update

lukaspj:
This update will allow any application to hook into T3D as a library and act as a scripting layer for it. With this update, you can write a Go layer, a C# layer, a Java layer or anything else you can think of. It is fairly easy to hook in.

The CInterface already exists in T3D, this just seeks to bring it up-to-date and solve the issues that have popped up from years of negligence.

I made this PR before it is ready to be merged, in the hopes that some comments on the changes I've made and the solutions I've done could improve my work.

E.g. the hook in to the CodeInterpreter is probably not done the cleanest way possible, or the changes I've done in EngineTypes.h is something I'm not too sure of.

Any questions are welcome.

PR #2226, opened 2018-03-19
Add initPersistFields to mPropertyTable

lukaspj:
Currently the ConsoleFunction `exportEngineAPIToXML` doesn't export properties properly.
This is due to a half-implemented Engine type system, where Console Classes define their properties in-between IMPLEMENT_CLASS and END_IMPLEMENT_CLASS.

This system was never put into use, and classes are implemented and ended right away directly in the IMPLEMENT macros, as seen here:
https://github.com/GarageGames/Torque3D/blob/development/Engine/source/console/consoleObject.h#L1171-L1173

`exportEngineAPIToXML` assumes the new system was taken into use, thus it never finds the properties as we still use `initPersistFields` to define our properties.

This change takes all the properties added in `initPersistFields` and adds them to the EnginePropertyTable, allowing `exportEngineAPIToXML` to work as intended.
This will help 3rd party applications analyze and provide auto-completion as well as code-generation for the T3D EngineAPI layer.
Improving the capabilities of 3rd party scripting tools.

Note:
Be aware that the EngineProperty type allows for more meta-information than what I'm filling into them.

The properties can be further marked as:
* Constant
* Transient 
* HideInInspectors

If there is any way to determine whether a field from initPersistFields is any of these 3, we should probably add that information to improve the quality of the data.


lukaspj:
Turns out this will need the fixed_tuple implementation in #2136, I thought it worked but I had my code-bases messed up.  I will close this and make a new one branched from #2136.

PR #2229, opened 2018-03-20
5 Button mouse support

OTHGMars:
SDL supports up to 5 buttons per mouse, T3D currently only supports 3. This commit adds event mappings for the additional buttons.


OTHGMars:
Withdrawing this PR. It will be resubmitted after or as part of SDL event refactoring.

PR #2233, opened 2018-03-29
Use a circular buffer for getReturnBuffer to prevent buffer cloberring

CouleeApps:
Try this in devhead:

    echo(1, 3);

Notice how `ConsoleValueRef::getStringValue` gets the same buffer as `echo` from `Con::getReturnValue` and causes this to echo `33`? Well it also causes other problems. This fixes that by defining a circular temporary buffer to replace `StringStack::getReturnBuffer`.

Also included is moving most of `StringStack`'s methods into the cpp file.

### DIFF
--- a/Engine/source/console/compiledEval.cpp
+++ b/Engine/source/console/compiledEval.cpp
@@ -41,6 +41,7 @@
 #include "core/frameAllocator.h"
 
 #include "console/codeInterpreter.h"
+#include "console/returnBuffer.h"
 
 #ifndef TORQUE_TGB_ONLY
 #include "materials/materialDefinition.h"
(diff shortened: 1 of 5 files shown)


jamesu:
Won't this ReturnBuffer still clobber in the following case?

	ReturnBuffer ret; // size: 4096
	ret.getBuffer(4096, 0); // mBufferSize = 4096, mStart = 0, returns &mBuffer[0]
	ret.getBuffer(1024); // mBufferSize = 4096, mStart = 1024, returns &mBuffer[0]

Wondering here if it might be a good idea to just make the ReturnBuffer grow until it gets reset in the CodeInterpreter, probably after a script function call (also taking into account nesting)...

CouleeApps:
It will, that's a good point. We may instead want to use a `FrameTemp<char>` but I haven't looked into them enough to know if that's a smart idea.

PR #2237, opened 2018-04-03
Small Cleanup to CodeInterpreter

JeffProgrammer:
Remove unused variables and cleanup precision warnings as dSprintf takes a U32 for the size of the buffer to use.

Honestly though, dSprintf should probably use dsize_t. But that's a bigger change down the road.

### DIFF
--- a/Engine/source/console/codeInterpreter.cpp
+++ b/Engine/source/console/codeInterpreter.cpp
@@ -426,12 +426,12 @@ exitLabel:
          }
          if (thisNamespace && thisNamespace->mName)
          {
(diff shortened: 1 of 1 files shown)


dottools:
`dSprintf` needs to be fixed to `size_t` instead of casting to that argument to `U32`. Perhaps @CouleeApps could look into this? :)

All data sizes should use `size_t`, or `ssize_t` for signed values if necessary. Especially now that these days 32bit operating systems are starting to not be a thing anymore just as 16bit computing died out in the 1990s. Network protocol handling is obviously an exception to this rule when compacted data values are used in packet processing.

rextimmy:
It's not just dSprintf that needs to use dsize_t, there are heaps of classes that needs this done. E.G Vector, String & more.

PR #2241, opened 2018-04-06
WIP: Added some missing field info to the EngineApi interface

lukaspj:
This probably doesn't have any effect on how the engine works, but it is nice for documentation and engine API export purposes.

### DIFF
--- a/Engine/source/core/util/tVector.h
+++ b/Engine/source/core/util/tVector.h
@@ -63,11 +63,12 @@ extern bool VectorResize(U32 *aSize, U32 *aCount, void **arrayPtr, U32 newCount,
 template<class T>
 class Vector
 {
-  protected:
+public:
    U32 mElementCount; ///< Number of elements currently in the Vector.
    U32 mArraySize;    ///< Number of elements allocated for the Vector.
    T*  mArray;        ///< Pointer to the Vector elements.
 
+protected:
 #ifdef TORQUE_DEBUG_GUARD
    const char* mFileAssociation;
    U32         mLineAssociation;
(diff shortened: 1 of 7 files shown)


Engine/source/core/util/tVector.h
```diff
@@ -63,11 +63,12 @@ extern bool VectorResize(U32 *aSize, U32 *aCount, void **arrayPtr, U32 newCount,
 template<class T>
 class Vector
 {
-  protected:
+public:
```
JeffProgrammer, line 66: quite dangerous, completely breaks encapsulation...

lukaspj: Yes I forgot I made that change in there, wanted to bring it up and ask if there were another way to do it.. Maybe moving the Implement Struct to somewhere associated with Vector?

lukaspj:
Added WIP tag, until we can figure out how to handle Vector stuff.

Engine/source/core/util/tVector.h
lukaspj: An alternative solution is hard-coding the offsets:
```c++
IMPLEMENT_STRUCT( Vector< bool >,
   BoolVector,,
   "" )
   {"elementCount", "", 1, TYPE<U32>(), 0},
   {"arraySize", "", 1, TYPE<U32>(), 4},
   {"array", "", 1, TYPE<bool*>(), 8},
   //FIELD(mElementCount, elementCount, 1, "")
   //FIELD(mArraySize, arraySize, 1, "")
   //FIELD(mArray, elements, 1, "")
END_IMPLEMENT_STRUCT;
```

lukaspj: Or the offsets could be computed and defined in tVector.h, but that sort of breaks encapsulation as well.

lukaspj:
Structs that have access permission issues:

* Vectors (IntVector, FloatVector, BoolVector)
* UUID
* MatrixF

lukaspj:
Closed in favor of #2366

PR #2245, opened 2018-04-17
Clean-up uses of ConsoleFunction etc.

lukaspj:
This PR Converts any use of ConsoleFunction or ConsoleMethod to DefineEngineFunction and DefineEngineMethod respectively, except for variadic functions or irregularly ordered default arguments.

Furthermore, this PR eliminates all use of DefineConsoleFunction, DefineConsoleMethod and DefineConsoleStaticMethod.

These 3 macros, were made as a temporary solution for functions that should exist in the Console system but not the EngineAPI.
While there exists examples of functions that should be removed, if the Console system was removed, this no longer seems relevant, as removal of the Console system is not in the road-map, so I think we should begin to settle for a Console / EngineAPI hybrid. Removing DefineConsoleFunction etc, is a step towards that.


lukaspj:
It would be possible to completely eliminate the use of ConsoleStaticMethod, but it would take a change to the scripts, as PfxVis is not a valid ClassName. Someone shorthanded the namespace for PostEffectVis, and DefineEngineStaticMethod doesn't allow for just typing whatever as the className.

lukaspj:
Note the latest commit, it completely eliminates the use of ConsoleFunction and ConsoleMethod and replaces it with new EngineAPI/Console hybrid functions (like DefineEngineMethod).

This won't break anything, but that is my decision on a design direction.

All the other commits are following up on past design decisions, the last commit is my solution to walk further down the design-direction.

PR #2247, opened 2018-04-21
Corrects a problem with the D3D11 texture lock/unlock mechanism

rextimmy:
As per the title. If you need some quick code to test this with see [here](https://hastebin.com/fexocuzudu.php) . This will produce a green texture 1024x1024 that will have a 512x512 blue box in the dead center of it.


OTHGMars:
I've tested this PR with a wide variety of rectangle sizes and offsets and it has produced correct results in every test. It performs identically to the opengl version now.

Areloch:
With that last fix, looks good now.

PR #2252, opened 2018-05-09
It's almost imposible to change direction of wind. Reseting mCurrentT…

pacomont:
…arget will allow to  rotate properly.


Areloch:
Oh, nice!

PR #2256, opened 2018-05-13
Adds support for SDL_Joystick and SDL_GameController

OTHGMars:
SDL 2.0.8 added functions that are exposed to script by this PR, so this PR relies on #2254.


OTHGMars:
Withdrawing this PR. It will be resubmitted after or as part of SDL event refactoring.

PR #2262, opened 2018-07-04
member var conversion error that oddly didn't crop up till mac testing.

Azaezel:


Azaezel:
followup to #2236. sorry about that folks.

PR #2270, opened 2018-08-07
micro patch to the nativefiledialogues library to mirror file type name

Azaezel:
folks with 'hide extensions for known file types' on windows weren't seeing any entries in thier drop-down lists for file types.


Azaezel:
see also https://forums.torque3d.org/viewtopic.php?f=12&t=1381 for what put us on to this. with special thanks to @OTHGMars for the assist

PR #2273, opened 2018-09-14
ExtendedMove packet cleanup

OTHGMars:
The `ExtendedMove::pack()` function will write 360 bits of data into every move packet even if no tracked devices are present. This can be verified by building a project with the TORQUE_EXTENDED_MOVE flag set, placing a breakpoint at https://github.com/GarageGames/Torque3D/blob/development/Engine/source/T3D/gameBase/extended/extendedMove.cpp#L148, launching the game with no tracked devices connected and stepping through to see every value get sent in full. All of the values being compared from the current move have already been clamped/unclamped and because of floating-point errors `0.0f != UNCLAMPPOS(CLAMPPOS( 0.0f ))`. This PR creates a clamped copy of the NullExtendedMove to use for comparisons in the pack function so the full extended tracking data is only sent for devices that are connected and tracking. The initial rotation values have been changed to reflect the change from QuatF to AngAxisF and prevent the `mDegToRad()` call [here](https://github.com/GarageGames/Torque3D/blob/development/Engine/source/T3D/gameBase/extended/extendedMoveList.cpp#L78) from causing an empty move to differ from the default.

### DIFF
--- a/Engine/source/T3D/gameBase/extended/extendedMove.cpp
+++ b/Engine/source/T3D/gameBase/extended/extendedMove.cpp
@@ -27,6 +27,8 @@ F32 ExtendedMoveManager::mRotAA[ExtendedMove::MaxPositionsRotations] = { 1, };
 
 F32 ExtendedMoveManager::mPosScale = 2.0f;
 
+static ExtendedMove ClampedNullExtendedMove;
+
 void ExtendedMoveManager::init()
 {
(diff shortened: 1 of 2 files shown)


jamesu:
Looks good

OTHGMars:
This PR is being withdrawn in favor of an ExtendedMove cleanup that will remove Euler rotations and mask off inactive devices eliminating the need for this.

PR #2276, opened 2018-09-28
filter out pixel shader normalmap calcs when not in deferred mode.

Azaezel:


dottools:
Uh oh, tabs vs spaces issue. ;)

PR #2278, opened 2018-10-21
OpenALEFFECT

marauder2k7:
Adds Reverb zone abilities to openal more work will be done to this to add the other effects but also to cleanup and comment the code so its more obvious what everything does


dottools:
Nice, thanks for the contribution.

marauder2k7:
no problem :) it is in no way pretty lol but the next update of this i will cleanup the code and comment it so people know whats going on and have more effects in it :)

PR #2281, opened 2018-10-29
OpenALEffects

marauder2k7:
Ground work, this commit enables reverb zones and opens the slots for
other effects to be added later. Tested using the full template. Build
your openal dll and your full.exe use one of the environment presets and
it will work


marauder2k7:
i should add this also gives the option for exponential falloff

Areloch:
Seems to all work, did note a typo with "PresetFroest" presumably that's supposed to be PresetForest?

Also, if you could correct the indentations on the SFXEnvironment definitions, and the typo, i think it's good to go :)

marauder2k7:
yeah it was supposed to be preset forest lol sorry rushing to test it out was a bit excited.

Areloch (approved):

wcb88:
bump
edit: works for me

PR #2283, opened 2018-11-23
Adds handlers for sdl focus events.

OTHGMars:


dottools:
Yikes, how was this missed? Good fix.

PR #2286, opened 2019-01-01
Nfd update

Azaezel:


Areloch:
With the default settings, looks to work fine on Ubuntu, no issues noted.
Also looks to work fine in Win. Need to test on the mac before merge-in to be sure it's all kosher(And give a run of the zenity mode to ensure that plays nice as at least an alternate mode)

jmdejong:
I think you need to add nfd_gtk.c to the blacklist when using zenity, otherwise it will just run gtk anyways.

There's a few bugs in nfd zenity:
https://github.com/GarageGames/Torque3D/blob/8304168365138c4d70d1f2a0c0ae1accfa945554/Engine/lib/nativeFileDialogs/nfd_zenity.c#L95
This should be `char* prefix = "--filename=";` instead. See also https://github.com/mlabbe/nativefiledialog/pull/60


https://github.com/GarageGames/Torque3D/blob/8304168365138c4d70d1f2a0c0ae1accfa945554/Engine/lib/nativeFileDialogs/simple_exec.h#L124
waitpid will always error (return -1) when the handler for SIGCHLD is set to SIG_IGN (which is the default).
https://stackoverflow.com/questions/22681021/waitpid-with-execl-used-in-child-returns-1-with-echild/24599676#24599676
This can be fixed by setting an empty handler for SIGCHLD.
See https://github.com/mlabbe/nativefiledialog/pull/61 for my proposed solution.
Alternatively the assertion could be removed.

PR #2292, opened 2019-01-07
afxRenderHighlightMgr: account for hardware skinning

Azaezel:


dottools:
Tabs vs spaces makes baby Jesus cry.

PR #2294, opened 2019-01-08
Adds features to GuiInputCtrl

OTHGMars:
This commit adds three new features to the GuiInputCtrl. All three default to off, so it is fully backward compatible with existing scripts. The new options are:
sendAxisEvents – If true, the control will generate onAxisEvent() callbacks for all axis events. This is useful for binding joystick/controller axes to game actions.
sendBreakEvents – If true, the control will generate onInputEvent() callbacks for SI_BREAK events for all keys and buttons. By default the callback is only triggered for break events on modifier keys.
SendModifierEvents – If true SI_MAKE events for modifier keys will generate callbacks. By default, only the break events are sent for modifier keys.


Azaezel:
checks out

PR #2295, opened 2019-01-08
Add a .editorconfig file

Bloodknight:
This file automatically overrides any default editor settings, in a variety of different development environments and editors. This is built into VS2017 but you can visit the site https://editorconfig.org/#download and hopefully download a plugin for your favourite editor.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# https://editorconfig.org/
+
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


Bloodknight:
More info about VS2017 support here https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2017

PR #2296, opened 2019-01-10
Fix SDL Input::getKeyCode on software keyboard layouts

CouleeApps:
Input::getAscii goes Torque keyCode -> SDL Scancode -> SDL Keycode -> SDL ascii key name
Input::getKeycode goes SDL ascii key name -> SDL Scancode -> Torque keyCode

This mismatch makes software keyboard layouts behave incorrectly in different places. For example, if you bind a key to an ActionMap and it will activate with a different button than specified.


Azaezel:
checks out

PR #2289, opened 2019-01-03
Travis script

jmdejong:
Tried to write a travis script that compiles Torque according to http://wiki.torque3d.org/coder:compiling-in-linux#toc11

This is still WIP. I will probably start a new branch/PR once I've got it working properly (unless it works on the first try), but for now I need a PR to have travis check my code.

PR #2211, opened 2018-02-27
Continous Integration for Torque3D

andijh92:
Hi,

Here are scripts for continous integration with appveyor.com and travis.com. These services are gratis for libre software projects. The CI is full featured and documented here ( http://wiki.torque3d.org/coder:continuous-integration ). You can see generated doxygen doc here ( https://andijh92.github.io/Torque3D/doc/dev/scriptReference/index.html ) and here ( https://andijh92.github.io/Torque3D/doc/dev/engineReference/index.html ). Compiling OSX doesn't work yet and I don't have osx to debug it. Maybe I'll fix it in the future.

Regards, Andy

PR #2298, opened 2019-01-10
Travis Compile

jmdejong:
This is essentially the same as #2289 but with a cleaner git history.

It follows the compilation steps from http://wiki.torque3d.org/coder:compiling-in-linux#toc11

This checks both gcc and clang on linux, but only clang on mac

Because there are so many debug warnings they are suppressed in a very hacky way:
Make is run first with error output ignored.
Then an incremental build is run which has the error output, but will only build the files that errored the first time (and will error again).
There is probably a better way to ignore the warnings in cmake, but I'm a bit hesitant to change the cmake files and I couldn't find a way to do it with command line arguments.

Only after writing this I found out that there is a CI pull request already: https://github.com/GarageGames/Torque3D/pull/2211
That older PR probably has more features (including running a windows build on appveyor, and I believe it runs the tests too) but this script is simpler.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,42 @@
 language: cpp
-compiler:
-  - clang
-  - gcc
+
+dist: xenial
+
+matrix:
+  include:
+    - os: osx
+      compiler: clang
+      env: if CXXFLAGS="-fgnu-inline-asm -fasm-blocks"
+    - os: linux
+      compiler: gcc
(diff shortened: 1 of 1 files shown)


Areloch:
We'll go ahead and roll this in for now(as mentioned, it's simpler) and can look to bringing over aspects of #2211 next to fully flesh out our CI setup.

PR #2299, opened 2019-01-10
Sqlite Console refactor, 

Bloodknight:
Also rename param to make tooltip function more friendly.

look tron, no tabs, in fact a few fell off :p


dottools:
Heh. It wasn't about tabs being a problem, mixing tabs and spaces indention is.

PR #2305, opened 2019-01-29
corrects a copy-corruption flaw with GuiSwatchButtonCtrl::onMouseDragged

Azaezel:
pretty sure the offending leak-line creeps in with
 https://github.com/GarageGames/Torque3D/blob/561f010f2e6411d8253d23f0cfcff794e81f60bf/Templates/BaseGame/game/tools/base/utils/swatchButtons.ed.cs#L34
comparative release/debug results in
string gridBitmap = "\c0"
vs
string gridBitmap = "ic\c1Gl\c0"
and the relavent initpersistfields entry of TypeString points to
https://github.com/GarageGames/Torque3D/blob/561f010f2e6411d8253d23f0cfcff794e81f60bf/Engine/source/console/consoleTypes.h#L71
which is likely causing malformation


Azaezel:
as a followup, if folks want to preemptively eyeball similar initpersistfields entries along the same lines, a list based on a wip-branch can be found https://gist.github.com/Azaezel/617204d2ffef388a2dd63ed849fd5466 , otherwise we'll deal with it on a case by case basis

PR #2307, opened 2019-01-31
nextfreemask does nothing for proximity mines as there are no subclas…

Azaezel:
…ses on this chain, but does bump it's enum value  over the limit with the new addition to sceneobject. so killed it.


Jacobtarango1996 (commented):
H

PR #2313, opened 2019-02-13
Switches to absolute position for mouse tracking.

OTHGMars:
The SI_XAXIS and SI_YAXIS (mouse absolute position SI_MAKE) events are sent whenever the mouse moves while the platform cursor is visible.


Azaezel:
verified functional

PR #2319, opened 2019-02-23
Initial implementation of the Scene object

Areloch:
Initial implementation of the Scene object for handling scenes/levels in a more consistent and deliberate way.


Azaezel:
Looks like incorporation of this one'll necessitate shifting https://github.com/GarageGames/Torque3D/blob/development/Templates/Modules/FPSGameplay/levels/EmptyLevel.mis#L2 to ```new Scene(MissionGroup) {``` for load support

PR #2323, opened 2019-02-23
Sanity check for calling getFieldValue

Areloch:
Sanity check so calling getFieldValue on a blank fieldName doesn't cause a crash.


dottools:
Not legal for [const] char *. Instead do:
```diff
    // Parse out index if the field is given in the form of 'name[index]'.

    const U32 nameLen = dStrlen( fieldName );
+   if (nameLen == 0)
+      return "";
    if( fieldName[ nameLen - 1 ] == ']' )
```

Areloch:
Oh, derp, yeah good call there

PR #2336, opened 2019-03-11
Fixes artifacts in Cloud Layer.

wcb88:
Cloud Layer uses normal. This applies appropriate profile for normal texture.


Azaezel:
can verify this one.

PR #2350, opened 2019-04-02
Temporary fix for the geometry feed in spotlights

ghost:
Done by Azaezel
Spotlights were breaking depending of the angle and distance after a certain radius.

Preview:
Sp1
sp2


dottools:
Why do you keep repeating the same PR over and over again?

ghost:
> Why do you keep repeating the same PR over and over again?

I'm not used to do anything using Github, the bot keeps throwing an error and i don't know where is the problem
Sorry for all of  that heh

Issue #2353, opened 2019-04-07
Openal-soft Broke Linux


practicing01:
1) Missing includes when compiling.
2) Crash on application start after successful compile.

practicing01:
This seems to fix it: https://github.com/practicing01/Torque3D/commit/5c5bfdd0a7019f86ed4e8fde64342a3d1621462c

Alexmac22347:
same issue with me, missing openal.h headers, segmentation fault when I get it to compile.
Will try out the fix, thanks for the info!

Areloch:
Merged in #2362

PR #2362, opened 2019-07-02
OpenAL efx fixes for macOS/Linux

rextimmy:
Fixes for using the new efx code with macOS and Linux. This PR also changes linux to use the T3D supplied openal-soft instead of the system supplied one, also changes macOS to use the T3D openal-soft instead of the now deprecated apple opeanl version (that didn't support efx extensions anyway). All 3 platforms now use the same openal-soft build.


rextimmy:
Oh yeah should give credit to @practicing01 , he created the original linux efx fix see #2353

PR #2365, opened 2019-08-01
WIP: Csharp torque3d work

lukaspj:


dottools:
Why are you changing AFX fields? They're not normal properties as for most of them each time they're changed it's pushing the field referenced datablock value to a dynamic array to queue up multiple identical elements of the same type. Hence why they're prefixed with `add` to indicate it's adding an additional element to the respective effect/action type.

lukaspj:
I realize that there is a reason they are pre-fixed.
The issue is that there is already a method called "addEffect":
https://github.com/GarageGames/Torque3D/blob/c131084a0d767415aa3b0f7bb39e95d04d11d179/Engine/source/afx/ce/afxPhraseEffect.cpp#L300
https://github.com/GarageGames/Torque3D/blob/c131084a0d767415aa3b0f7bb39e95d04d11d179/Engine/source/afx/ce/afxPhraseEffect.cpp#L186

Which one of these do I get if I do `%obj.addEffect` ?
Personally I believe these fields are a weird sort of ugly magic and I don't agree with it. But if they have to exist, they should be named something else. I don't really care what, but it shouldn't overlap with the method.

Azaezel:
based on a review of the docs and samples, would say keep the varname, change the method to push<foo>

lukaspj:
Closed in favor of #2366 
Because I changed branch.

PR #2367, opened 2019-08-13
Action Adventure module

wcb88:
This is basically ready to merge. It works if Action Adventure Kit (AAK) is merged. However, the FPSGameplay module will have to be modified because if you merge AAK and try to use that module then when you go into third person there is very odd horizontal clamping on the camera and weird tank controls for A and D.


wcb88:
Oops. Nevermind.

PR #2368, opened 2019-08-13
add Action Adventure module to BaseGame template

wcb88:
This is basically ready to merge. It works if Action Adventure Kit (AAK) is merged. However, the FPSGameplay module will have to be modified because if you merge AAK and try to use that module then when you go into third person there is very odd horizontal clamping on the camera and weird tank controls for A and D.


wcb88:
And by "ready to be merged" I mean not at all ready due to that FPSGameplay problem ;)

Also, there are missing textures. But it's the basic structure for an AAK module at any rate.

wcb88:
Needs fly cam because can't get around when in world editor. Also sound is broken. Need to also make sure SFXSpaces are still working.

wcb88:
Also need to fix it so it respawns on death and also need to fix the texture issues on the climbable surfaces

wcb88:
I've started on a AAK module for torque 4 BaseGame so I will no longer be updating this one.

wcb88:
https://github.com/wcbx/torque4-aak-module

PR #2382, opened 2019-11-22
Remove trailing whitespaces from simObject.cpp and fix pointer format

stsrc:
These are just small changes. 

Trailing whitespaces removal was automatically done in vim
Fixing pointer format in getDebugInfo() helps debugging under gdb.


dottools:
I hate editors that don't right-trim source files. Visual Studio has historically been bad at doing this. Caused by pressing enter and it auto-aligns new line but didn't trim white-space that carried over that was trailing before the cursor position of previous line.

